PR #12, opened 2015-08-05
Prevent CMAKE_Fortran_COMPILER_VERSION from becoming undefined

raullaasner:
- [x] `CMAKE_Fortran_COMPILER_VERSION` was added to CACHE to prevent it from becoming [unset after the second configuring](http://stackoverflow.com/questions/31805891/cmake-fortran-compiler-version-becomes-unset-after-second-configuring). Previously, this variable was sometimes automatically unset which in turn affected variables such as `opencoarrays_aware_compiler`. I reported a [bug](http://public.kitware.com/Bug/view.php?id=15684) about this so let's wait for a little while for the response.
- [x] The variable `CMAKE_Fortran_COMPILER_VERSION` is present starting with CMake 3.3 and therefore I think it's better to write the conditional not as `(CMAKE_VERSION VERSION_GREATER 3.2.3)` but `(NOT CMAKE_VERSION VERSION_LESS 3.3)`. Similarly, at line 45, if the intention is to test whether GCC is at least version 5, it would be better to have `(NOT CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 5)`. In [other places](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/extensions/CMakeLists.txt), I can see conditionals such as `(CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER 4.9.2)` (note that 4.9.3 has been released). If you agree with me, then I'd like to replace `VERSION_GREATER` with `NOT ... VERSION_LESS` and make the necessary changes everywhere.


raullaasner:
Regarding the first issue - the bug has been confirmed and will be fixed starting with CMake 3.3.1. A temporary fix is required only until then.

sourceryinstitute:
Thanks for the pull request.  I just merged it into the master branch.  My apologies for the delayed review.  I returned to work on Wednesday after a 10 days of business travel and vacation travel.

PR #20, opened 2015-11-11
Fix the link_args variable inside the caf script.

milancurcic:
This commit introduces a fix to src/mpi/CMakeLists.txt which contains
a rule that creates the link_args shell variable that is used inside
the caf script. Before the fix, the caf script contains a line:

link_args=-fcoarray=lib -lcaf_mpi

which produces an error when running caf:

$ caf -w
/opt/opencoarrays/gcc-5.1.1/mpich-3.1.4/bin/caf: line 37: -lcaf_mpi: command not found
caf wraps CAFC=mpif90

link_args needs to be enclosed in double quotes:

link_args="-fcoarray=lib -lcaf_mpi"

so that caf can work as expected.

Signed-off-by: milancurcic <caomaco@gmail.com>


milancurcic:
Resolved by commit https://github.com/sourceryinstitute/opencoarrays/commit/ef5b51103c19446e6a1b23a7aa1acf8496ddd6c7.

Issue #28, opened 2015-12-01
test 6, "allocate_as_barier_proc", failing


zbeekman:
See #27 for some context/backstory

Is there a known issue with test 6 "allocate_as_brier_proc" when using OpenMPI and GFortran 5.2? I know from the [known issues](https://github.com/sourceryinstitute/opencoarrays/blob/master/STATUS.md#known-issues) that "co_reduce" won't work until GFortran 5.3/6.1 but what's the story with this other test?

zbeekman:
This question is intended primarily for @afanfa, I believe

zbeekman:
relevant to https://github.com/Homebrew/homebrew/pull/46547

afanfa:
Hi, I cannot reproduce the issue with allocate_as_brier_proc. I'm using gcc-5.2, opencoarrays-1.2.0, mpich-3.1.4, cmake-3.1.3 on Ubuntu. Can you post the error you get?

zbeekman:
OK let me see what's going on. I was about to say with confidence that it happens with OpenMPI on Mac OS X all the time, however, I just tried it with OpenMPI (and clang for mpicc) and it didn't happen. It was happening with gcc 5.2 for mpicc when building on the travis-ci homebrew test servers. Let me dig around some more and get back to you.

zbeekman:
this is maddening.... it seems that there is something from my environment that allows the build and test (by hand) with brewed GCC 5.2, Clang, OpenMPI to work, but when building and testing under the restricted environment of Homebrew this doesn't work.

The output is just simply: "Test failed." It doesn't give any reason. No segfaults or anything like that.

zbeekman:
OK upon looking at the definition of the unit test, and reading the output more carefully, it says:
```
test 6
      Start  6: allocate_as_barrier_proc

6: Test command: /usr/local/bin/mpiexec "-np" "2" "/tmp/opencoarrays20151202-44375-14o6csk/opencoarrays/build/src/tests/unit/init_register/allocate_as_barrier_proc"
6: Test timeout computed to be: 9.99988e+06
6:  Test failed.           1
 6/21 Test  #6: allocate_as_barrier_proc .........***Failed  Required regular expression not found.Regex=[Test passed.
]  1.07 sec
```

So [here](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/init_register/allocate_as_barrier_proc.f90#L11) `me /= 1` but on the following line it does equal 1...

[...]

Issue #27, opened 2015-12-01
failing tests


zbeekman:
There is probably an issue with my MPICH and OpenMPI installations, but I'm going to ask this question anyway:

Is there any reason why when building and testing OpenCoarrays tests 6, "allocate_as_barier_proc" and 18, "co_reduce" are failing yet all the other tests pass? I think/hope I've gotten a working [OCA Homebrew Formula](https://github.com/Homebrew/homebrew/pull/46547), but I'm also have been having trouble building OCA from source and via the Homebrew formula on my local mac. I can't tell if this is a problem with my toolchain or a larger issue. The CI tests haven't finished yet for the OCA formula, so we'll see what happens there; my first attempt [failed because cmake couldn't find a working mpif90 compiler](https://travis-ci.org/Homebrew/homebrew/jobs/94237438#L660), which was odd.

sourceryinstitute:
> On Dec 1, 2015, at 2:00 PM, Izaak Beekman <notifications@github.com> wrote:
> 
> There is probably an issue with my MPICH and OpenMPI installations, but I'm going to ask this question anyway:
> 
> Is there any reason why when building and testing OpenCoarrays tests 6, "allocate_as_barier_proc" and 18, "co_reduce" are failing yet all the other tests pass? 
> 
> 

See bullet 4 at https://github.com/sourceryinstitute/opencoarrays/blob/master/STATUS.md#compiler-issues-gnu.  If your compiler version is lower than the (pre-release) 5.3.0 or the (pre-release) 6.0.0, then co_reduce will fail.  Hopefully 5.3.0 will release relative soon and 6.0.0 will release next Spring. 

D

zbeekman:
What about the test "allocate_as_barrier_proc"?

rouson:
> On Dec 1, 2015, at 2:11 PM, Izaak Beekman <notifications@github.com> wrote:
> 
> What about the test "allocate_as_barrier_proc"?
> 
> 
Good question.  If you haven’t already, please create an issue report and hopefully Alessandro can investigate this.

D

zbeekman:
OK, making good progress on the Homebrew formula, I have it more or less building except for failing these tests. I'll close this issue for now, since it has been superseded by #28, for which we're just waiting for @afanfa's input. This seems to be the only thing breaking the build now that I have made some restrictions on compilers (telling homebrew it can't use clang or llvm).

PR #43, opened 2015-12-07
CMake opt flag fix

zbeekman:
This PR makes it so that user selected CMake build types don't add optimization flags to GFortran. This is due to a GFortran bug which @afanfa has apparently already submitted a patch for. I don't know what it's status is, however.

@afanfa, do you have a link to the issue tracker or some other means of monitoring the bug?

For more back story see #27, #28 and Homebrew/homebrew#46547 and also https://groups.google.com/forum/#!msg/opencoarrays/RZOwwYTqG80/46S9eL696dgJ


afanfa:
You can track the review process here: https://groups.google.com/forum/m/#!topic/opencoarrays/IOx0RFLJEr0

zbeekman:
Also, @rouson and @afanfa,
Do you think I should create an advanced cache variable to allow users to turn of my `-O0` override? Perhaps some people want to live dangerously and prefer faster code even if it could be incorrect? Thoughts?

zbeekman:
Some more tweaks and changes are required as per discussion with @rouson... Will try to get to them soon.

zbeekman:
This was rebased onto the changes being made to master to bring it closer to branch-1.x

zbeekman:
Fixed as #71

PR #47, opened 2015-12-12
Travis ci

zbeekman:
Pull request to enable Travis-CI continuous integration testing on the master branch.

I'm intentionally doing this before attempting to reconcile some issues I discussed with @rouson about master and branch-1.0.0 being out of sync, so that I/we have the added assurance of the CI tests when I start to tackle that.

Also, I disabled the `coarray_navier_stokes` test but ONLY when running under Travis-CI since it is producing an error that can't be produced elsewhere. 

If anyone can determine the source of that issue, seen [here](https://travis-ci.org/zbeekman/opencoarrays/jobs/96364146#L688) we can certainly turn this test back on for Travis.


zbeekman:
Closing to reattempt getting travis to build this PR

Issue #52, opened 2015-12-13
Rectify differences between `master` and `branch-1.0.0`


zbeekman:
Ensure the only differences between the 1.x release branch, `branch-1.0.0` and `master` are intentional and resolve any unintended conflicts.

Some differences are clearly intentional, like the mpi_cuda work on master, and the install script modifications on the release branch (which likely may be safely absorbed into master, but I don't know if @rouson wants me to attempt this or if he would prefer doing it himself).

```diff
 CAF_API.txt => CAF_ABI.txt                          |  159 ++--
 CMakeLists.txt                                      |   73 +-
 COPYRIGHT-BSD3                                      |   18 +-
 GETTING_STARTED.md                                  |    2 +-

branch-1.0.0 wins for all these:
 doc/dependency_tree/opencoarrays-tree.txt           |    9 +
 install.sh                                          |  807 +++++++++++++++----
 install_prerequisites/build                         |  130 +++-
 install_prerequisites/stack.sh                      |  209 +++++

This was updated in mid-November in preparation for SC15.  Go with whichever branch has a commit just before the November 15 tutorial (presumably master branch):
 src/cuda_mpi/Makefile                               |   21 -
 src/cuda_mpi/mpi_caf.c                              | 2517 ------------------------------------------------------------
 src/extensions/caf-foot                             |    1 +
 src/extensions/opencoarrays.F90                     |  221 +-----
 src/make.inc                                        |    2 +-
 src/mpi/mpi_caf.c                                   |   69 +-
 src/tests/integration/gpu/Makefile                  |   27 -
 src/tests/integration/gpu/co_dot.f90                |  152 ----
 src/tests/integration/gpu/kernelCaller.cu           |  176 -----
 src/tests/unit/extensions/CMakeLists.txt            |    6 +-
 src/tests/unit/extensions/this_image_num_images.f90 |   62 --
 src/tests/unit/simple/CMakeLists.txt                |    4 +-
 20 files changed, 1113 insertions(+), 3552 deletions(-)
```

rouson:
If you feel comfortable doing so, please go ahead and move install.h to the master branch and move the entire install_prerequisites/ directory to master as well, replacing the version of that directory on master. Nothing in master's version of install.sh needs to be saved and nothing in master's install_prerequisites/ directory needs to be saved. Of course,  all of this needs to be tested, but it seems that should be automatic now that we have Travis-CI. :)

zbeekman:
[...]

PR #57, opened 2015-12-14
Part I of II issue 52 re-syncronize master and 1.x branch

zbeekman:
This is the first part of a two part PR to bring master and branch-1.0.0 back in sync.

This branch proposes changes to be applied to the master branch to incorporate mainly build system improvements spearheaded by @rouson. I am fairly confident of these changes but would greatly appreciate code review by @rouson and @afanfa.

The goal of these two PRs is to reduce the changeset between master and branch-1.0.0 to the minimal set of changes required such that:
 - All build system improvements and bug fixes on branch-1.0.0 are applied to master
 - All new features, that are unstable or inappropriate for the 1.x release line are only on master.
 - Version strings and software stack dependencies remain as relaxed as possible on the 1.x branch, but are incremented, where needed, on master to support new features.
This is further discussed in #52.

The github change-browser is going to have some trouble minimizing the set of diffs due to the diverged history. 

Once I post the Part II PR, (which will represent proposed changes to the 1.x branch) I will also post a link with some sort of diff between the proposed master and branch-1.x branch)

:pray:


rouson:
Comments:

Let's bump the minimum gfortran version to 5.3, which is required to eliminate the co_reduce test failure.  Of course, I know that will trigger a GCC build on almost all platforms, but I think we ought to just go all the way and get as much right as we can.  We should simultaneously change the default version that install.sh builds to 5.3 as well.  That will make installations more consistent across platforms.  Until now, the script would build the 6.0.0 trunk if the minimum wasn't met.  I can help with these changes if you like.

CAF_ABI.txt should be renamed CAF_ABI.md to get the Markdown formatting correctly when the file is accessed via a browser on github.com.

zbeekman:
> Let's bump the minimum gfortran version to 5.3, which is required to eliminate the co_reduce test failure. Of course, I know that will trigger a GCC build on almost all platforms, but I think we ought to just go all the way and get as much right as we can. We should simultaneously change the default version that install.sh builds to 5.3 as well. That will make installations more consistent across platforms. Until now, the script would build the 6.0.0 trunk if the minimum wasn't met. I can help with these changes if you like.

Are the problems with 5.1 and 5.2 and collectives due to bugs or lack of support?

I ask because I'm of two minds about this suggestion. On the one hand, rebuilding and reinstalling GCC is a PITA, on the other hand, using gfortran 5.1 and 5.2 will potentially produce incorrect behavior. If it is a compiler bug, then I am inclined to bump the minimum version to 5.3 as you suggest. If it is lack of support, then a vociferous warning in the CMakeLists.txt when using gfortran 5.1 and 5.2 would suffice, IMO, while changing the install script to use 5.3 by default, and be the minimum required there...

> CAF_ABI.txt should be renamed CAF_ABI.md to get the Markdown formatting correctly when the file is accessed via a browser on github.com.

:+1:, I'll rename it.

zbeekman:
@afanfa It looks like [you have work on your fork of master relevant to using a helper thread](https://github.com/sourceryinstitute/opencoarrays/compare/master...afanfa:master) that has not been merged into the sourceryinstitute/opencoarrays master branch yet. Do you want me to wait until you do so before I merge these commits, or are you confident in you git abilities to reapply your work on top of these commits?

rouson:
The problem with 5.1 and 5.2 was a bug in the compiler side. Alessandro fixed it but the patch missed making it into the 5.2 release by hours if I recall correctly.  I'm fine with freezing at 5.3 as the minimum for some time if that helps -- probably for as long as the 5 branch is maintained, which I imagine will be for another year or more. Keep in mind that our 1.0.0 release was less than 5 months ago and the first release of the installation script was just over 1 month ago.   I think it's ok that the requirements are still evolving a bit.  Once we settle on 5.3 as the minimum, we can hold steady and keep the requirements stable for the foreseeable future.

afanfa:
@zbeekman, if you could wait until tomorrow I would appreciate it. I also need to add code that have not been committed on my fork.

PR #59, opened 2015-12-14
Part II of II issue 52 re-syncronize master and 1.x branch

zbeekman:
This is the second part of a two part PR to bring master and branch-1.0.0 back in sync. (See also #57)

This branch proposes changes to be applied to the 1.x branch to be incorporated into the master branch. These are mostly cosmetic changes, and typo fixes & line ending fixes to reduce the number of differences between the branches.  I am fairly confident of these changes but would greatly appreciate code review by @rouson and @afanfa.

The goal of these two PRs is to reduce the changeset between master and branch-1.0.0 to the minimal set of changes required such that:
 - All build system improvements and bug fixes on branch-1.0.0 are applied to master
 - All new features, that are unstable or inappropriate for the 1.x release line are only on master.
 - Version strings and software stack dependencies remain as relaxed as possible on the 1.x branch, but are incremented, where needed, on master to support new features.
This is further discussed in #52.

The github change-browser is going to have some trouble minimizing the set of diffs due to the diverged history. Rectifying this and re-establishing a more recent common history will help git "do the right thing" when analyzing merges and PRs in the future due to its sophisticated merge strategy algorithm, using a common ancestor to help reduce conflicts.

:pray:


rouson:
Looks great -- thanks tons!

zbeekman:
@afanfa and @rouson:

It looks like sometimes the `allocate_as_barrier_proc` (test 6) is failing on Travis-CI under OS X *with* GFortran **5.0.3** and [MPICH installed from the "bottle" provided by Homebrew](src/tests/unit/extensions/). (This uses [Apple LLVM version 6.0 (clang-600.0.54) (based on LLVM 3.5svn)](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/97491206#L139)

Is it possible that the GFortran patch from #28 didn't completely fix the issue?

However, this behavior is not deterministic; sometimes it fails, other times it does not. It had failed for commit 0022426 here, but then I re-ran the CI-testing on the apple image, and, as you can see above, it passed.

afanfa:
The patch has been applied to gcc-6 (trunk) and gcc-5.4 (not yet released). So, it makes sense that the test fails with gcc-5.0.3. The "sometimes" is quite scary.

zbeekman:
ohhhhh, OK then. So I need to look at the CMake script and make sure I am disabling it for <= 5.3... I thought I had done that already, but maybe I had specified < 5.3 not <= 5.3 i.e. enabled for >= 5.4

Thanks for the clarification @afanfa

zbeekman:
OK, I also forgot that I hadn't merged #43 yet, as well. So I'm going to reopen #28 until #43 is merged, which I'll probably try to do after I reconfigure the branches for github flow.

PR #66, opened 2015-12-17
Merging async progress feature

afanfa:
Adding support for asynchronous progress. This feature is still under development and it has been disabled by default.


afanfa:
Same pull request merged as #67

PR #70, opened 2015-12-17
Fix sloppy EOL & EOF whitespace

zbeekman:
Apply the same white-space fixes to the 1.x branch


zbeekman:
OK, next commits will be fixing that darn error! (#28)

Issue #65, opened 2015-12-16
Sticky thread for binary prerequisite uploads


zbeekman:
## Current status

Please note, this is not a bug at all. This issue is being kept open to make it easier to find, until a better/cleaner solution is implemented. This thread is a place to upload binary assets needed by our CI testing infrastructure. In this way, github issues are being used as a bootleg CDN, of sorts. At some point I think that I could create a project, using a combo of https://nightli.es, https://travis-ci.org and https://bintray.com to provide a free service for building and uploading MPICH and OpenMPI Homebrew bottles using the latest GCC. This has the potential to speed up or enable a lot of MPI projects that want to use GCC over clang on OS X using CI testing with Travis-CI.
## Old Text

The MPI implementation for the Travis-CI builds on OS X cannot be built from source, and the only "bottled" installations use clang as the mpicc command. (Without verbose output the built times out after 10 minutes without anything being logged---i.e., quietly building MPI---and with verbose output the log length exceeds 50mb which also causes a failure.) This doesn't appear to impact the functionality of OCA, however, generating coverage data (#51) becomes nearly impossible due to some files requiring one version of gcov and other files requiring a different version. The codecov.io script is very good at "just do what I mean/want" and replacing it with manual coverage generation, collection, transmission etc. is just too daunting a task. (It requires a uniform version for gcov.) 

Unless @rouson has a better idea of a server to host binary files (inappropriate for inclusion in a git repository) I'm going to try using the Github issues file uploader mechanism as a content delivery network (CDN) to provide pre compiled/bottled homebrew MPICH and OpenMPI implementations to the Travis-CI OS X workers that are built against gcc-5. My only other thought is to host it at my UMIACS.UMD.EDU public FTP page. This should greatly speed the OS X builds as apposed to building from source, and allow testing of the GCC-5 gcc compiler, along with generating coverage info.

Here is a test bottle upload: [umobj-1.3.0.yosemite.bottle.tar.gz](https://github.com/sourceryinstitute/opencoarrays/files/64217/umobj-1.3.0.yosemite.bottle.tar.gz)

This bottle is completely unrelated to OCA, but _is_ a homebrew bottle.

zbeekman:
Closing issue to see if file is still accessible.

zbeekman:
[...]

Issue #51, opened 2015-12-13
Add code coverage analysis


zbeekman:
Use [codcov.io](https://codecov.io/github/sourceryinstitute/opencoarrays?ref=master) to enable tracking and checks for code test coverage. Generate coverage information for as much as we can, but only measure/report on coverage of the core API/ABI files written in C and Fortran, and tested during Travis-CI unit/integration testing.

Browser plugins are available [here](https://github.com/codecov/browser-extension) to overlay coverage info when browsing gitub (Chrome, FireFox, Opera)

Here is a short video on the browser extensions:
[![video][10]][11]

[10]: https://cloud.githubusercontent.com/assets/2041757/9814630/7c8847d0-585d-11e5-887b-b5a19170ef61.png "Watch how on YouTube"
[11]: https://www.youtube.com/watch?v=d6wJKODB8_g

zbeekman:
Got unified Mac and Linux coverage reporting working via #65 

See https://codecov.io/github/zbeekman/opencoarrays?view=all&ref=a53ac7e1da99f661fc2f8f8add17f95b793f4cf3

Note the menu item that says "unified". Here you can pull it down to look at the results for just one configuration.

rouson:
Thanks for this. It's a great idea. 

It suddenly strikes me that there are limitations to the code coverage concept (as you've already pointed out).  For example, I'm a big fan of the Fortran MERGE intrinsic, which I use to write one-line conditional assignments.  Hitting a statement like the following once only tests one branch of the inherent conditional:

function is_positive(x) result(which_am_i)
  real, intent(in) :: x
  logical :: i_am_positive
  i_am_positive = merge(.true.,.false.,x>0)
end subroutine

so even a code coverage that is nominally 100% might not actually test all possible branches in the code.  The above assignment could result in  either which_am_i having a value of .true. or .false. depending on the value of x.

D

zbeekman:
Yes, you are right. AND, even if a line of code was executed, there's no guarantee that the test was properly written to expose any defect that might be on that line. However, I think it can be a useful tool assess the state of general test coverage, and help make sure that contributors also write tests for their new code. It does give a macro assessment of the overall state of testing and QA.

PR #71, opened 2015-12-17
Issue 28 opt flag fix

### LINKED ISSUE #28: test 6, "allocate_as_barier_proc", failing

### LINKED ISSUE #31: Don't delete files generated at config time when `make clean`
I think this is caused only by the examples from scientific software design, like the burgers example.

**Problem description:** `make clean` followed by `make` fails because `compiler_capabilities.txt` is added to `additional_make_clean` property in CMake which tells make to remove the file, but it is generated to configure time, not compile time, and therefore should probably not be removed by `make clean`

### LINKED ISSUE #56: bad links in README.md
Some links have been broken, on both branches.

zbeekman:
- Fixes #28 : When CMake release types are passed to CMake, optimization flags are added to the gfortran compiler, which trigger bugs in GCC for GFortran < 5.4
 - Fixes #31 : Generate files at CMake config time in the binary tree not the source tree. Also don't delete them with `make clean`
 - Fixes #56 : A number of bad links existed in the README.md

Normally, once we're on the Github Flow workflow, some of these may have deserved their own branch (i.e., issue-28-gcc-flag-bug, issue-31-make-clean, issue-56-bad-links but in the interest of time and paying down technical debt, I have applied all of these on the same branch, in the same PR.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,9 @@ env:
     - MPICH_URL_HEAD="http://www.mpich.org/static/downloads/$MPICH_VER"
     - MPICH_URL_TAIL="mpich-${MPICH_VER}.tar.gz"
     - MPICH_DIR="$HOME/.local/usr/mpich"
-    - OSX_PACKAGES="gcc cmake mpich"
-    # homebrew bug requires MPI to be built from source, for now
+    - MPICH_BOT_URL_HEAD="https://github.com/sourceryinstitute/opencoarrays/files/64308/"
+    - MPICH_BOT_URL_TAIL="mpich-3.2.yosemite.bottle.1.tar.gz"
+    - OSX_PACKAGES="gcc cmake"
 
 matrix:
   include:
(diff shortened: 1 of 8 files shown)


zbeekman:
It seemed as though this patch was not fixing the issue. I've added some more debug output and force pushed (to my *private* branch) an updated version. Hopefully tests pass on the first try.

zbeekman:
I think I have fixed the issue with test 6, however right now Homebrew's CDN seems to be down, and as a consequence `brew install mpich` is trying to install from source which times out. As a work around, I'm also going to add in my fix for #51 because it grabs the bottles from Github, which acts as our CDN through uploading them to #65.

zbeekman:
Merging this even though, on occasion, #28 is still failing on OS X. If this continues, we can tell Travis-CI to ignore the outcome of the OS-X build; it will still do it, but failures there won't impact the build status...

PR #78, opened 2015-12-19
Update development branch readme etc. and OS X travis bug work around

zbeekman:


zbeekman:
Still learning about Github protected branches and PR merging. I think that the best tack is to to follow #77

PR #83, opened 2015-12-20
Enable install.sh to build CMake on OS X by using Apple LLVM

rouson:
Use the gcc and g++ that is installed in /usr/bin by XCode with its Command-Line Tools.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -545,12 +545,35 @@ find_or_install()
     else # permission granted to build
       printf "n\n"
 
-      if [[ -z "$CC" ]]; then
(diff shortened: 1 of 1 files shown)


zbeekman:
:+1: :ship: it! :zap:

zbeekman:
(Unless otherwise noted, comments like the one I made above indicate that I've read over the changes, and have no objections or comments on them.)

rouson:
Thanks for the review.  I'll commit within 24 hours of submission unless Alessandro weighs in with any concerns.

zbeekman:
Yup, sounds good.

PR #91, opened 2015-12-26
Bump version # to 1.2.3

rouson:
Both ./install.sh --version and caf --version output the VERSION argument from "project" statement in the top-level CMakeLIsts.txt file. Immediately before this commit, git clones show a number that matches the most recent release even though new commits have happened. This commit bumps the VERSION from 1.2.2 to 1.2.3-prerelease to distinguish the current development version from the most recent release and the upcoming release.


zbeekman:
:eyes: :+1:

PR #92, opened 2015-12-26
Eliminate inconsequential install error messages.

rouson:
The stack_new function invoked by install.sh outputs inconsequential
errors on OS X because of the lack of support for the -g flag in bash
"eval".  This commit pipes that output to /dev/null.


zbeekman:
:+1: from the :airplane:

rouson:
Nice!  Did you review this kn your phone?

zbeekman:
Yes

zbeekman:
I'm really confused as to what happened here, but now it looks like this PR was based on `devel` not master. The consequence of this, if it is true, is that `devel` and `master` are now basically the same, and the intentional differences have been added to master. 

I'll try to look at this first thing tomorrow, I just landed but have more traveling today.

rouson:
That makes two of us!  My goal was to merge the two branches that I created recently into master and into devel.  I believe that worked with one of them (the one that eliminates the stack.sh error messages) but not with the other (the one that bumps the version number). From the network graph, it appears that th latter one merged into master but not into devel.  Attempting to merge via GitHub's web interface leads to error messages that recommend specific command-line steps.  Those command-line steps lead to errors (namely the recommended "git checkout -b ..." step gives a "fatal" error because it apparently is trying to overwrite the existing branch).   My guess is that I am yet again at a point where the only reasonable solution is to do a fresh clone, but I'm not sure that even that will solve this one.  If I can't do the merge with a fresh clone, then I'll give up and wait until we can look at this together.  

It seems to be impossible to find a _simple_ work flow with git where the same steps work every time.

zbeekman:
I need to dig a bit deeper, but I'm guessing that one of the branches may have been based on  `devel` when the work started... Not sure though. Still waiting for my bag at JFK

rouson:
Now I think I've accomplished the goal of merging my two recent branches into both master and devel.  If I also somehow made devel and maser the same, that was unintentional.  Let's hope I haven't wreaked additional havoc.  If I also made devel and master the same, that was unintentional so my apologies. If you can figure out how to reverse that, please feel free to do so. I have one minor new feature I'd like to work on in the installation script soon, but it can wait a little while if that's best.

PR #99, opened 2016-01-08
Handle missing wget gracefully.

rouson:
The "build" script now detects a missing download mechanism (as designated in the $fetch variable) and prompts the user to download the relevant tar ball and restart the installation.  A few exit statements have also been updated to have unique exit codes. A minor edit in "install.sh" indicates to the user that the tar ball will be downloaded "if necessary", i.e., if it isn't already found in the expected location.

This has only been tested when fetch=wget.  It should be able to detect if another missing required download mechanism (e.g., svn or ftp) is missing, but it's possible the printed messages will be less clear in that case.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -534,7 +534,7 @@ find_or_install()
       echo "$this_script: Ready to build $package executable $executable in $package_install_path"
     fi
 
-    printf "$this_script: Ok to download, build, and install $package from source? (y/n) "
+    printf "$this_script: Ok to download (if necessary), build, and install $package from source? (y/n) "
     read proceed_with_build
 
     if [[ $proceed_with_build != "y" ]]; then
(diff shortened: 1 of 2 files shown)


zbeekman:
Branch based on `master`? :heavy_check_mark:
Branch merging into `master`? :heavy_check_mark:
Changes consistent with intended effect, commit message and PR description? :heavy_check_mark:

Looks good to me! :eyeglasses: :100:

szaghi:
@zbeekman what a beautiful commit comment!

zbeekman:
@rouson If you like I can merge this and then also apply it to `devel`

rouson:
Let's see if I can get it right this time.  :)

If I encounter any difficulties, we can do a Skype call anytime after 9:30 AM Pacific.

zbeekman:
sounds good :-)

I may be in the car, so an old fashioned phone call might have to suffice.

PR #93, opened 2015-12-26
Fix inconsequential stack sh errors

rouson:

PR #100, opened 2016-01-09
Handle missing wget gracefully

rouson:


zbeekman:
Please do NOT merge this. :heavy_multiplication_x: 

Ugggg merge commits. :frowning: 

I need to dig around some more. We may need to take some more drastic and aggressive actions to sort out the accidental merge from #93 or #92

rouson:
Agreed. I just finished writing a related message to you on gitter.im

zbeekman:
Alright, since it seems @afanfa has also been only working off of his private fork, https://github.com/afanfa/opencoarrays and because the accidental merge commit of old devel into master from PR #92 caused such a headache, I have simply rebased `devel` onto `master` at caae07634fd44d41d107f79507dea71af655593b and force pushed to `origin`. This has greatly simplified the network graph, and eliminated the merge commits that were causing such a headache. I think I have a less drastic plan of attack for if this happens again in the future, but since I'm re-writing some history now, I thought I would also take the opportunity to fixup/squash some commits together to get a cleaner commit history.

rouson:
FYI, I might as well capture my reasoning here:

Just in case it's ever useful, the [build](https://github.com/sourceryinstitute/opencoarrays/blob/master/install_prerequisites/build) script is capable of building several packages that we don't currently tell it to build automatically inside install.sh.  For example, it can build wget and m4 (I'll post a separate issue about m4 momentarily).  I thought about inserting logic into install.sh so that it would do something like "cd install_prerequisites && ./build wget" when wget is missing.  However, then one has to have ftp (or curl, etc.) to download wget and wget itself has one or more prerequisites that would have be downloaded as well (e.g., [pkg-config](http://www.freedesktop.org/wiki/Software/pkg-config/)).  Considering all of this, I decided that it wasn't worth the effort to try to have install.sh.  It seems simpler to just let the user know it's missing (in case they want to install it themselves) and give them the URL for each missing prerequisite so that can download it in a browser or whatever is their chosen download mechanism. As long as they put the specified tar ball in [install_prerequisites](https://github.com/sourceryinstitute/opencoarrays/tree/master/install_prerequisites),  [build](https://github.com/sourceryinstitute/opencoarrays/blob/master/install_prerequisites/build) will find it and not attempt to download it.

zbeekman:
I completely agree. After all, we can't account for every possible software configuration and at some point it becomes impractical to support the combinatorially explosive number of possible configurations, especially when we're already attempting to download any missing software for the user. Plus, it's not clear how one would automate the download of the download agent... that's just getting a little bit too meta.

rouson:
LOL. Great response.  I nee to learn how to insert emojis here.

zbeekman:
github emoji start with a `:`. So if you type `:+1:` that gets rendered as :+1:

After the initial `:` autocomplete starts suggesting emoji to insert :tada: :bow:

rouson:
:+1:

Issue #29, opened 2015-12-01
Document & clarify development model


zbeekman:
Hi Guys @sourceryinstitute, I'd love to help out some more, but I'm finding it hard to follow the branching/release/development model. It seems the two main branches, master and branch-1.0.0 [have diverged significantly](https://github.com/sourceryinstitute/opencoarrays/network).

I assume this is intentional, and you're following some kind of "git-flow" development model?

The development model should be documented somewhere and branches described as well as more detailed procedures for authoring PRs, like which branch they should be based on, and made against.

rouson:
Hi Zaak,

I wrote a rather lengthy reply and then accidentally hit a Safari bookmark that took me to another page and I lost everything I wrote. :(  I should get in the habit of writing responses in a separate document and then pasting them here.

I have to head home for dinner with the family but hope to reconstruct what I wrote later.

Damian

zbeekman:
Damian,
Sorry to hear you lost your hard work. No rush, just have a few minor things that I spotted and thought needed fixing.

rouson:
Ok, Take 2:

Your question nicely assumes that we have a branching/release/development model. :)  Your help in developing one is welcome.  Alessandro and I talk weekly so you could join one of our upcoming teleconferencesto discuss this if you like.  For now, I'll explain my thoughts and influences informally and why I'm comfortable with the divergence between the two branches. 

First, the seeming divergence is partly superficial divergence because I manually copy files from one branch to the other (almost always from the release branch to the master branch).  I do this only to avoid the complications of git merge.  Git for me is like democracy, which has been described as the worst form of government except for all the alternatives.  I think possibly some of my previous difficulties related to frequent rebasing.  My life seems so much simpler now that I've stopped rebasing.  "Hi. I'm Damian and I don't rebase." :D 

Second, the minor new features that have been added on the release branch (branch-1.0.0) have very little if any interaction with the core library.  Version 1.1.0 added scripts for building the OpenCoarrays prerequisites. Version 1.2.0 added the installation script, which will see some very significant improvements in the imminent 1.2.1 release.  It's very easy and safe to copy these sorts of new features over to the master branch, which is our pre-release 2.0.0 branch as I'll explain below.

The following have most influenced my thinking on branching, version numbering, and releasing:

1. The GCC approach (https://gcc.gnu.org/develop.html), especially the version numbering scheme they adopted with release 5.
2. Semantic Versioning (http://semver.org).
[...]

PR #105, opened 2016-01-11
Merge latest install logic including m4 dependency into `devel` branch

zbeekman:
Pick up PR #103 in `devel`

EDIT: I spoke too soon...
EDIT 2: My first comment was correct, but a "protected branch" setting was causing Github to suggest "updating" the PR branch (in this case `master`) by merging the base branch (`devel`) into it, _definitely **not**_ what we actually want to do... jeesh

Much simpler than merging `devel` into `add-m4-to-dependency-tree` and then `add-m4-dependency-tree` into `devel`. :bulb:

Git is also smart enough to know, since `devel` is based on `master` but we intentionally made changes like bumping the project version number to 2.0.0 etc. not to change that back to 1.2.3 etc. :tada:

I really need to address #29

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,6 +9,10 @@
 *.F90 text
 *.md text
 *.txt text
+*.sh text
+install_prerequisites/build text
+*.cu text
+*.x64 text
 
 # Denote all files that are truly binary and should not be modified.
 *.mod binary
@@ -22,3 +26,4 @@
 .travis.yml export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
+developer-scripts export-ignore
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ set ( CMAKE_BUILD_TYPE "Release"
 set_property ( CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS ${CMAKE_CONFIGURATION_TYPES} )
 
 #Name project and specify source languages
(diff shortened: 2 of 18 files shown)


zbeekman:
EDIT: I spoke too soon... I'll investigate on the command line, but please DO NOT MERGE yet...

zbeekman:
OK, I think I found the source of the issue whereby Github was disallowing merging of `master` into `devel` because `master` was "outdated": I had [designated `devel` as a protected branch](https://github.com/sourceryinstitute/opencoarrays/settings/branches) (as is denoted by the shield/keyhole icon on the ["branches" page](https://github.com/sourceryinstitute/opencoarrays/branches)) and had [required even administrators to pass status checks](https://github.com/sourceryinstitute/opencoarrays/settings/branches/devel). For some reason, in addition to the Travis-CI and CLA-assistant checks, Github tries to make sure that the merge is able to be a fast-forward merge (i.e. all changes being merged are applied starting from the "base" branch) even though merging PRs like this do not cause merge conflicts.

In essence, [my first comment, now struck through](https://github.com/sourceryinstitute/opencoarrays/pull/105#issue-126053543) was correct, but I had turned on a Github setting that I thought would help keep us out of trouble, that actually had the opposite effect.

zbeekman:
Taking a look at the ["files changed" tab](https://github.com/sourceryinstitute/opencoarrays/pull/105/files) displays the desired changes to be replayed on top of `devel`. This looks good to merge, to me, but I'll wait at least until tomorrow AM before doing so.

PR #108, opened 2016-01-22
Merge build script improvements.

rouson:


zbeekman:
You merged before anyone signed off!

(We need to allow this, admins to always be able to merge, otherwise, it tries to tell us that the branch is out of date and needs updating, when in fact we're doing the right thing....)

zbeekman:
LGTM!

zbeekman:
Approved

rouson:
Darn.  I was thinking that this one was past the 24-hour waiting period so it made sense to merge it.  I guess I should have first added a comment with one of the expected regular expressions.  Correct?

zbeekman:
Yes exactly.. but to be fair we're just getting started with this feature, and I haven't documented/explained it in full yet.Those regexp are actually inspired by [Github's actual code review process](http://scottchacon.com/2011/08/31/github-flow.html) at least, as it was in 2011. (See "#5 - Merge only after pull request review") Feel free to edit .pullapprove.yml to add more.

Also, it's really too bad, that the automated checks have that odd requirement to make sure that the branches "aren't out of date" when applied to repository administrators. It's very strange to me why they have this. Maybe I should shoot them a note, that it at least needs to be documented more clearly.

zbeekman:
And, I'm going to write up a development guide soon to document the workfow etc. My basic thinking is, in general, for non trivial (i.e. documentation changes, etc.) only merge after 1) Someone else has approved the PR, or 2) You approve the PR >= 24 hrs after pushing the last changes to it or opening it.

zbeekman:
I think this will make clearer the conditions under which the PR was merged. Those big green comment badges stand out, and all the comments have time stamps.

BTW, you can change things to require, for example, that you need a certain number of reviewers' approval, or a certain number from different groups, etc. Might come in handy down the road if SI/OC grows a lot.

rouson:
Ok.  Let's go over this in next week's Skype call.

Issue #101, opened 2016-01-09
refactor .travis.yml


zbeekman:
Break out shell script from `.travis.yml` into dedicated scripts to improve readability and debugging.

See also #80

zbeekman:
The macOS builds always take forever to start because the resources are always highly contended. The GCC 6 and GCC 7 builds should be done from the same macOS instance in the job matrix, and we should investigate whether we could run all the tests (GCC 6, 7 and GCC 7 via install.sh) from the same instance. This would require that the total time spent in all the tests cumulatively is less than the timeout limit.

Another advantage of combining tests into a single entry in the build matrix is that they will share a lot of the installation overhead, so some of the time will be amortized because few dependency installations will take place.

PR #110, opened 2016-01-23
Expand test matrix to include install.sh script and `build`

zbeekman:
Missing from this PR, i.e., still under development:

 - Clever managing which dependencies are pre-loaded through more expedient Travis-CI logic, when build scripts have not been touched, but building as much as we can when modified in the commit range being tested by Travis-CI
 - Testing car wrapper scripts after OpenCoarrays has been installed (This isn't done in any tests yet)

Additionally, `.travis.yml` is getting unwieldy and could benefit from being broken up into one or more shell scripts ( #101 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,38 @@ env:
     - MPICH_BOT_URL_HEAD="https://github.com/sourceryinstitute/opencoarrays/files/64308/"
     - MPICH_BOT_URL_TAIL="mpich-3.2.yosemite.bottle.1.tar.gz"
     - OSX_PACKAGES="gcc cmake"
-    - BUILD_TYPE="CodeCoverage"
 
 matrix:
   include:
     - os: osx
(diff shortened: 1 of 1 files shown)


zbeekman:
Note, that the indiscriminate "build everything you can" approach causes the tests to run for close to 30 minutes instead of under two. That's why it will be important to examine the commit range to see if `install.sh` or anything in `install_prerequisites/*` has changed and avoid building CMake and MPICH in those cases.

zbeekman:
:zap:

rouson:
Your message has the "| pullapprove | approve |" buttons but the code-review/pullaprove still shows as pending.  ????

zbeekman:
@rouson, I don't know, but maybe? I got in touch with pull approve support to see what they have to say.

zbeekman:
:+1: Will a second try work?

Issue #119, opened 2016-01-26
Source-to-source translation with OFP


zbeekman:
For the 2.x major release
-  ~~@zbeekman: Add OFP as git submodule~~
- [ ] @rouson & @rasmussn : Implement source-to-source magic

zbeekman:
OFP added as git submodule in PR #127.

To get the OFP code into your working copy: `git submodule update --init --recursive`

To clone a new copy of OpenCoarrays and get the OFP code, `git clone --recursive https://github.com/sourceryinstitute/opencoarrays`

zbeekman:
The wrong OFP project was added as a submodule. The correct one is https://github.com/OpenFortranProject/ofp-sdf

Furthermore, we don't anticipate needed to modify any of the OFP code, and hope to use it as a stand alone dependency, so there is no need to use it as a submodule.

zbeekman:
PR #137 Is the first step in this direction.

zbeekman:
@rouson Should we close this as "won't fix" for right now? It seems like this might not be progressing anytime soon.

zbeekman:
@rouson I'm going to close this due to inactivity. Feel free to reopen if you disagree.

PR #127, opened 2016-01-31
Add OFP as a git submodule

zbeekman:
See #119 as well.

 I have updated the git setup/config scripts to help ease working with
 submodules. I suggest re-running the git setup developer script.

 Submodules are a somewhat complicated beast, but I really think for our usage,
 pulling in OFP via submodule sounds like the best route. If we change our minds
 once we start hacking with OFP, we'll be able to investigate other alternative
 integration strategies. (Subtree merge, committing code directly, scripted external
 dependency, etc.)

 To begin reading up: https://git-scm.com/book/en/v2/Git-Tools-Submodules

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -27,4 +27,5 @@ install_prerequisites/build text
 .pullapprove.yml export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
+.gitmodules export-ignore
 developer-scripts export-ignore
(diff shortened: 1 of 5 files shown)


zbeekman:
No idea why pull approve thinks this has been approved...

zbeekman:
LGTM! #shipit

PR #128, opened 2016-02-01
Add OFP git submodule to devel

zbeekman:


zbeekman:
:zap:

zbeekman:
Merging because this is the `devel` contribution corresponding to #127 

Come to think of it, maybe I should not have pulled #127 into master, but only here into devel... it is listed as a 2.x milestone. We can discuss on Tuesday's call.

PR #133, opened 2016-02-02
Fix issue #123 build ftp fails on linux

### LINKED ISSUE #123: ftp fails on Linux
The current ftp command inside [build](https://github.com/sourceryinstitute/opencoarrays/blob/master/install_prerequisites/build) works on OS X but fails on Linux. Possible fixes: (1) revert to wget or (2) use a more involved multi-line ftp such as explained [here](http://www.stratigery.com/scripting.ftp.html)

rouson:
This fixes #123 and improves the setup.sh file that install.sh creates.  When install.sh builds prerequisites from source, "source setup.sh" sets LD_LIBRARY_PATH correctly for both 32- and 64-bit builds on Linux and OS X and it adds gfortran/gcc/g++ to PATH and correctly detects when to add m4 to the PATH.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -312,11 +312,11 @@ find_or_install()
       export CC=$package_install_path/bin/gcc
       export CXX=$package_install_path/bin/g++
       if [[ -z "$LD_LIBRARY_PATH" ]]; then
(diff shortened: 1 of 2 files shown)


rouson:
LGTM

zbeekman:
LGTM too

PR #135, opened 2016-02-04
Update copyright year for files touched in 2016

### LINKED ISSUE #130: Copyright  2015 --> 2016 update
A lot of references to 2015 need to be updated now that its 2016. I have posted a gist of all of them, found using [ag](https://github.com/ggreer/the_silver_searcher) a faster version of [ack](http://beyondgrep.com/). If you download the file in the gist linked below, you can open it in Emacs and then `M-x compilation-mode` to go visit each file and line mentioned. YMMV with other editors.

https://gist.github.com/zbeekman/a09504020f6689e5a43d

zbeekman:
Fixes #130

Find files with `2015` and have been modified since December 31, 2015:

```
git grep --name-only 2015 $(git log --pretty=format:%H --since="12/31/2015") | \
    cut -d : -f 2 | sort | uniq | \
    while read line; do \
       grep -Hn 2015 $line; \
    done > 2015-refs.ack
```

Then open `2015-refs.ack` in Emacs and `M-x compilation-mode` to jump to the file and line easily

### DIFF
--- a/COPYRIGHT-BSD3
+++ b/COPYRIGHT-BSD3
@@ -4,8 +4,8 @@
     interface, and compiler wrapper.
 
  OpenCoarrays is distributed under the OSI-approved BSD 3-clause License:
- Copyright (c) 2015, Sourcery, Inc.
- Copyright (c) 2015, Sourcery Institute
+ Copyright (c) 2015-2016, Sourcery, Inc.
+ Copyright (c) 2015-2016, Sourcery Institute
  All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/doc/dependency_tree/gitkeep.sh
+++ b/doc/dependency_tree/gitkeep.sh
@@ -14,8 +14,8 @@ use_case=\
     command 'tree opencoarrays'."
 #
 # OpenCoarrays is distributed under the OSI-approved BSD 3-clause License:
-# Copyright (c) 2015, Sourcery, Inc.
-# Copyright (c) 2015, Sourcery Institute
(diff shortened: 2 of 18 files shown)


zbeekman:
If I get a code review/:+1: from @rouson I'll merge tonight... Otherwise I'll wait until tomorrow

rouson:
Holy cow, batman!  There's no way I would have figured this out myself. Nice work.  :+1:

zbeekman:
@rouson I'm counting that as a 
LGTM

(Note that the regex all match the line start, so that quoting via email etc. doesn't introduce an accidental approval)

rouson:
@zbeekman, I thought we had multiple options for approving, including :+1: and LGTM.  Is that not the case?  And am I understanding you that the approval has to come at the beginning of the first line?

zbeekman:
There are multiple approval patterns *but* the approval pattern must also *start* at the beginning of the line. :shipit: <-- invalid
LGTM! Awesome extra comment <-- valid
:zap: <-- valid even though it's not on the first line...

I'm working on getting to that developer guide/manual.... someday in the not so distant future, I hope.

zbeekman:
Here is the regex: `approve_regex: '^([Aa]pproved|:\+1:|:shipit:|:ship: it|LGTM|[Ll]ooks good to me|:zap:|:boom:|:clap:|:bow:)'` Please note the leading `^` (matches the beginning of a line)  The point of this is so that when replying by email (and potentially quoting someone else response) you don't accidentally get flagged as approving the PR when someone else's comment containing one of the approval keywords is quoted.

rouson:
Makes sense.

PR #138, opened 2016-02-06
CMake cleanup #58

### LINKED ISSUE #58: reduce build system redundancy
It seems that there are three CMakeListst.txt files that check compiler version etc. to set `opencoarrays_aware_compiler`. The top level CMakeLists.txt does this, then it happens again in the `src/mpi/CMakeLists.txt` and `/src/tests/unit/extensions/CMakeLists.txt`. Is there a reason this value can't simply be inherited from the top level CMakeLists.txt? This will ease the burden of maintaining which upstream software is compatible with which feature...

zbeekman:
Resolves #58, mostly. Still some stuff left to do concerning cleaning up install directories, install logic, and improving/fixing the CMake project export capabilities. (This last feature is CMake's answer to pkg-config---It tells other CMake projects that want to use OpenCoarrays, what linker & compiler flags they need, where OC is installed, etc.)

 A number of extraneous CMake files and syntax were removed. Enhancements include:
 1. Use the "CMake way" to query system processor and OS
 2. Check if the compiler is OCA *once* in the top level CMakeLists.txt and then just query the variable, `opencoarrays_aware_compiler` indicating if the compiler is OCA
 3. Don't hack and distribute intrinsic CMake module: "CHECK_Fortran_SOURCE_COMPILES", use `CMAKE_REQUIRED_FLAGS` to pass `-fcoarray=single` to the introspection fortran compiler invocations.
 4. Fix bug wherein CMake is exporting the package to tell other CMake software how to find and build against OpenCoarrays, but it was always telling this software that it is V1.0 not the current version.


zbeekman:
LGTM

PR #145, opened 2016-02-09
Update devel with CMake cleanup from master

zbeekman:
Pull in latest CMake cleanup changes


zbeekman:
:boom: looks good to me!

PR #149, opened 2016-02-18
Bump ver.# for release. Delete redundant copyright

rouson:


zbeekman:
LGTM :zap:

zbeekman:
@rouson I hope you don't mind; I didn't realize that you had been working on drafting release notes, and ended up [drafting my own](https://github.com/sourceryinstitute/opencoarrays/releases/tag/untagged-952ced37e1047ba63220). I then deleted what I thought was still my "this is how to create a draft release" example, without looking carefully enough. However I had the releases page open in another browser tab, and noticed (too late) that you had started to draft real release notes in the draft I had just deleted. I think my release notes have the same content, plus some additional notes on changes and fixes. Please note that the Travis-CI build badge will show passing once the 1.3.4 tag is created and pushed to GitHub.com

Here is [a view showing commits and files changed since 1.3.2](https://github.com/sourceryinstitute/opencoarrays/compare/1.3.2...master)

Here are all the issues closed since 1.3.2: https://github.com/sourceryinstitute/opencoarrays/issues?utf8=✓&q=is%3Aissue+closed%3A%3E2016-01-29+

Here are all the merged PRs since 1.3.2: https://github.com/sourceryinstitute/opencoarrays/pulls?utf8=✓&q=merged%3A%3E2016-01-29+base%3Amaster+

rouson:
No problem on the mix-up. One question arises in looking at the view showing commits nd files changed since 1.3.2: it appears CheckFortranSourceCompiles.cmake was deleted. Is this correct?  If so, please explain. That file came from a suggestion by Brad King if I recall directly.  The idea is to detect actual compiler capabilities rather than to really simply on version numbers.

zbeekman:
@rouson:

> No problem on the mix-up.

Phew... sorry about that once again :sweat_smile: 

> One question arises in looking at the view showing commits nd files changed since 1.3.2: it appears CheckFortranSourceCompiles.cmake was deleted. Is this correct? If so, please explain. That file came from a suggestion by Brad King if I recall directly. The idea is to detect actual compiler capabilities rather than to really simply on version numbers.

Yes, CheckFortranSourceCompiles.cmake was deleted. It is [a module __included__ with CMake](https://cmake.org/cmake/help/v3.5/module/CheckFortranSourceCompiles.html). By duplicating it, we were essentially clobbering CMake's own version with ours. CMake maintains this file, and we would miss any future improvements to it by maintaining a parallel copy. As can be seen on [line 93 of the top level CMakeLists.txt diff](https://github.com/sourceryinstitute/opencoarrays/compare/1.3.2...master#diff-af3b638bc2a3e6c650974192a53c7291R93) I have replaced the usage of our version with using CMake's builtin version. Our version was slightly different than the intrinsic `CheckFortranSourceCompiles.cmake` in that an additional flag was passed to the compiler: However, CMake provides a means of setting up/altering the default flags passed to CheckFortranSourceCompiles. This is accomplished by [setting CMAKE_REQUIRED_FLAGS](https://github.com/sourceryinstitute/opencoarrays/compare/1.3.2...master#diff-af3b638bc2a3e6c650974192a53c7291R91) before calling the macro.

I completely agree with @bradking that system introspection via CMake is far better than trying to check versions and manage a model of features in each compiler/library/system/etc. and hope to find other places to simplify the build system and improve its robustness through introspection.

zbeekman:
> One question arises in looking at the view showing commits nd files changed since 1.3.2: it appears CheckFortranSourceCompiles.cmake was deleted. Is this correct? If so, please explain. That file came from a suggestion by Brad King if I recall directly. The idea is to detect actual compiler capabilities rather than to really simply on version numbers.

I revisited this code, after noticing a small problem with my changes: The intrinsic CMake module writes a .F file, whereas the modified version we had in the repo writes a .F90 version. As a result, this introspective test was failing. However, at least on `master`---I still need to check `devel`---nothing was being done with the info from the introspection, so it did't matter whether or not it passed or failed. If we're not using the result of this test on `devel` either, I'm going to completely remove it. In the mean time I have a "fix" (there was never really a problem...) so that this test will behave as expected once again.

zbeekman:
An unnoticed and unfortunate side effect of this PR, namely the deletion of `COPYRIGHT-BSD3`, was that [`install.sh` started silently erroring](https://travis-ci.org/zbeekman/opencoarrays/jobs/112699852#L499). I have a fix for this in a PR I am working up with numerous other changes.

rouson:
CheckFortranSourceCompiles.cmake will become increasingly important as we start to support non-GNU compilers. If you remove it from master, please make sure it stays on devel.

zbeekman:
:+1:

rouson:
> On Feb 29, 2016, at 3:10 PM, Izaak Beekman <notifications@github.com> wrote:
> 
> An unnoticed and unfortunate side effect of this PR, namely the deletion of COPYRIGHT-BSD3, was that install.sh started silently erroring <https://travis-ci.org/zbeekman/opencoarrays/jobs/112699852#L499>. I have a fix for this in a PR I am working up with numerous other changes.
> 

I assume the fix is just replacing COPYRIGHT-BSD3 with LICENSE.  Let’s plan to release tomorrow. This is a critical bug, given that it prevents the writing of the setup.sh file without which user’s won’t have a simple mechanism for defining necessary environment variables.

D

PR #156, opened 2016-03-04
Fix small CMake introspection issue

zbeekman:
- Ensure that CheckFortranSourceCompiles can compile test/introspection code which is in F90 not f77
 - Make sure `install.sh` returns proper exit code... A bunch of the other scripts need to be examined for this defect too....


zbeekman:
LGTM

PR #158, opened 2016-03-08
Corrected install.sh script's writing of setup.sh

rouson:
Removed an extraneous/erroneous "export PATH=.." that install.sh writes near the tope of setup.sh.
Corrected the instructions for how to use setup.sh so that sudo privileges won't be required.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -52,6 +52,10 @@
 #    Alternatively, check the dependency stacks for the package before entering the
 #    main conditional blocks in find_or_install.
 #
+ 
+# Make sure we really exit with correct status when piping function output
+set -o pipefail
+
 # __________ Process command-line arguments and environment variables _____________
 
 this_script=`basename $0`
--- a/install.sh
+++ b/install.sh
@@ -54,9 +54,6 @@
 #
 # __________ Process command-line arguments and environment variables _____________
 
-# Make sure we really exit with correct status when piping function output
-set -o pipefail
-
 this_script=`basename $0`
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -54,9 +54,6 @@
 #
 # __________ Process command-line arguments and environment variables _____________
 
-# Make sure we really exit with correct status when piping function output
-set -o pipefail
-
```
zbeekman: @rouson is this being intentionally removed? This is needed to ensure that the correct exit status is returned when the script fails.

I'm thinking maybe you need to merge master into your branch (or git rebase master) since this was changed in  a recent commit.

rouson: That deletion was a mistake.  I'll cancel this pull request, commit a correction, and submit a new pull request.

I thought about merging master into the caffeinate-opencoarrays branch, but I didn't know how to do it in a way that preserves the version number on the branch so I cherry-picked this commit, assuming that would accomplish the same thing as merging.  Apparently it's not quite the same, given that the network graph doesn't depict the commit from master merging into the branch.

zbeekman: I think if you request the merge via a github pull request, then if the "files changed view" shows something that looks wrong, a manual `git merge --no-ff --no-commit master` should git you to a place where you can either resolve existing conflicts, or if errors are found upon inspecting the current state of the code, you can make changes and `git add` then `git commit # this should complete the merge commit... message should be pre-populated with Merge master into caffeinated...`

PR #159, opened 2016-03-08
Fix setup sh script

rouson:


zbeekman:
This is an example of where an interactive rebase to squash these two commits together would make sense... I'd be happy to do it myself or we can do it on our call.

rouson:
You're welcome to do it.  I certainly thought of it and would have done it if I remembered how, but the benefit didn't seem to outweigh the cost. Nonetheless, I'm not against the idea.  Feel free to squash away. :)

zbeekman:
Once tests pass, should be ready to merge

zbeekman:
:zap:

Issue #160, opened 2016-03-11
STOP normal termination: missing OUTPUT_UNIT output


mexas:
$ cat term.f90
integer :: i[*], j, img
  img = this_image()
  i = img
  if ( img .eq. 1 ) stop "img 1 cannot continue"
  do j = 1 , 100000000
    i = int( atan( real( j ) ) )
  end do
  write (*,*) "img=", img, "i=", i
end

$ caf -o term.ox -Werror -O2 term.f90

$ cafrun -np 5 term.ox
STOP img 1 cannot continue

The output from other images is missing.
Should be something like this:

img 1 cannot continue
 img=           3 i=           1
 img=           2 i=           1
 img=           4 i=           1
 img=           5 i=           1

This has been confirmed as incorrect behaviour at:
https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=comp-fortran-90;f35859ec.1603

Anton

afanfa:
Anton, thanks for reporting this. I'll fix it during the weekend.

Alessandro

afanfa:
Hi,
I've noticed that the STOP statement is translated into an invocation to a function not implemented in OpenCoarrays. This function is commonly used for serial execution and does not have any knowledge about coarrays.
I need to modify GFortran to distinguish coarray and non-coarray cases and invoke the right function accordingly. Obviously, in the coarray case, the program will invoke the function implemented into OpenCoarrays (to do).

afanfa:
Hi,

I've got a working patch for GFortran and OpenCoarrays about this bug. I'll submit them tomorrow.

Cheers,
Alessandro

rouson:
:clap:

afanfa:
The GFortran patch has been applied to trunk and gcc-5-branch.

afanfa:
Fixed

zbeekman:
@mexas I just wanted to confirm that this has solved your problem. We are issuing a new release today with these changes---since builds of OpenCoarrays with gcc-trunk or gcc-5-branch (GCC > 5.3) will have a build failure with previous releases of OpenCoarrays ( <= 1.3.8)

mexas:
I don't see any change.

$ which caf
/panfs/panasas01/mech/mexas/OpenCoarrays-1.4.0/opencoarrays-installation/bin/caf
$ which cafrun
/panfs/panasas01/mech/mexas/OpenCoarrays-1.4.0/opencoarrays-installation/bin/cafrun
[...]

PR #162, opened 2016-03-27
Bugfix STOP

afanfa:
Fixes bug #160 about STOP.


rouson:
:+1:

afanfa:
Could you please apply this patch to the devel branch as well?

Issue #154, opened 2016-03-03
Non-sudo OFP installation


rouson:
The caffeinate-opencoarrays branch now contains a first-draft [install-ofp.sh](https://github.com/sourceryinstitute/opencoarrays/blob/caffeinate-opencoarrays/install_prerequisites/install-ofp.sh) script for installing the Open Fortran Parser (OFP) and its prerequisites aterm, sf2-bundle, and Stratego/XT.  The script currently installs only on OS X and only by downloading and moving binaries into /opt because one or more binaries contains explicit references to the /opt directory.  

A future version of the OFP installation script should allow for building from OFP and its prerequisites from source in order to facilitate
1. Installation of OFP and its prerequisites into directories other than /opt and
2. Installation on platforms other than OS X. 

A preliminary draft of a script for building the OFP prerequisites from source is in  [installStrategoXT.sh](https://github.com/sourceryinstitute/opencoarrays/blob/caffeinate-opencoarrays/install_prerequisites/installStrategoXT.sh), which currently builds aterm successfully but fails on building sf2-bundle on OS X and Linux.  Once this script can be fixed, building OFP is trivial and as detailed [here](https://github.com/sourceryinstitute/opencoarrays/blob/caffeinate-opencoarrays/install_prerequisites/install-ofp.sh#L197).

zbeekman:
> Installation on platforms other than Linux.

Did you mean other than OS X?

rouson:
Sent from my iPhone

> On Mar 3, 2016, at 7:56 PM, Izaak Beekman <notifications@github.com> wrote:
> 
> Installation on platforms other than Linux.
> 
> Did you mean other than OS X?
> 
Yes.

zbeekman:
I updated your comment to reflect this

rouson:
@rasmussn, I'll remove this issue from blocking the 2.0 release targeted for May 30 and assume we'll work on it afterward.

scrasmussen:
Hi, is anyone working on a StrategoXT installation script for Linux users? If no one is and you guys don't mind I'd like to help with that. It'll target the non-binary package installation of aterm, sdf2-bundle, and StrategoXT since I have experience with that problem. I'll attempt to code in a similar style like you guys have done in [install-ofp.sh](https://github.com/sourceryinstitute/opencoarrays/blob/master/prerequisites/install-ofp.sh) so as to keep things consistent.

zbeekman:
[...]

Issue #118, opened 2016-01-26
Finalize CUDA MPI


zbeekman:
for 2.x release

 - [x] additional testing

rouson:
@afanfa, is there anything left to do on this or is this feature ready for release once we merge devel into master?

zbeekman:
Just waiting on release notes from @afanfa

afanfa:
The CUDA support provided by OpenCoarrays is based on the function registernc. This function simply registers the host memory stored during the normal allocation into the CUDA memory address space. This simple allocation delegates all the data transfers from/to host/gpu to the CUDA runtime.

PR #163, opened 2016-04-03
Pull in `stop` bug fix and other recent changes on `master`

zbeekman:
`devel` has fallen a little bit behind `master`...

Let's get them back in sync

 - #160 and #162 
 - #156 
 - #159


zbeekman:
This will require some manual command line merging. I also need to be extra careful since I'm a bit rusty on the current state of `devel`

zbeekman:
LGTM

rouson:
@zbeekman thanks for working on this.  My guess is devel will be dormant for a bit until we are close to the [2.0.0](https://github.com/sourceryinstitute/opencoarrays/milestones/2.x%20release) release, for which I've set a tentative target date of May 30.  The most important requirements for 2.0.0 in my view is the integration of OFP as marked by issue #154 and #119.  A close second would be #118, but that has a slightly lower priority because (1) a much broader set of similar functionality will be available when we enable CAFe after the OFP integration, (2) @afanfa won't generally be available for OpenCoarrays work after starting his postdoc (except for working on failed-image support, which is the subject of the postdoc), and (3) I'm pursuing funding related to #118 and it makes sense to await the funding decision before proceeding.  We'll see how many of the other milestone issues we can resolve but even just finishing the OFP integration is a big enough deal for the release -- assuming the [1.4.0](https://github.com/sourceryinstitute/opencoarrays/milestones/1.4.0%20release) release milestone issues have been closed before we approach the 2.0.0.  I've set a tentative date for 1.4.0 release of April 30.

zbeekman:
Sounds great!

PR #168, opened 2016-04-13
Add install script boilerplate

rouson:
The new scripts contributed in this merge will be useful when we're ready to take more steps toward adding TAU installation as an option in the OpenCoarrays installation process so it makes sense to merge these into the ptpkg branch that was created for work on the TAU installation script.  In particular, it probably makes sense to replace the ptpkg branch's install-tau.sh with a copy of this pull request's install-ofp.sh branch and then to refactor the new install-tau.sh so that it wraps ptkpkg instead of the current installation steps.  If for any reason that path doesn't work out, then the TAU installation might be accomplished by instead wrapping the build.sh script in this merge request.


rouson:
Closing and will resubmit a new pull request with squashed commits.

PR #169, opened 2016-04-13
Add install script boilerplate

rouson:
This merge contains the first fully functional Open Fortran Parser (OFP) installation script: install-ofp.sh. When no arguments are passed, the script downloads and installs the aterm, sdf2-bundle, and strategoxt prerequisites in OS X binary form in in /opt, which requires that the user have sudo privileges. The script then installs OFP in the subdirectory prerequisites/installations and finally the script builds the parse table required for parsing Fortran 2008.   Executing "./install-ofp.sh -h" shows the options the script accepts.  The script detects the identity of the operating system.  Future versions of the script will use that information to install binaries suitable for other operating systems or to build the prerequisites from source. 

This merge also does a tiny bit of housekeeping: it renames "install_prerequisites" to just "prerequisites" both for typing simplicity and to indicate that the directory is the default location of installed prerequisites rather than just the location for the scripts that build and install prerequisites.


rouson:
Closing and will resubmit a new pull request with squashed commits.

PR #175, opened 2016-04-26
Update caffeinate-opencoarrays branch: OFP install option

rouson:
This completes the refactoring of install.sh so that it now uses bash3boilerplate to achieve more sophisticated argument parsing, including the option to install OFP by issuing the command 
```
./install.sh -p ofp
```
which currently installs binary files that work only on OS X and only when the OFP prerequisite packages aterm, sdf2-bundle, and strategoxt are placed in /opt.  To see a list of other installation options, execute
```
./install.sh -h
```


zbeekman:
It may make more sense to merge master into caffeinated-opencoarrays after the boiler plate stuff is merged into master...

rouson:
Sounds good.  I'll close this pull request and open a new one after boilerplate-install-sh has been merged into master.

Issue #150, opened 2016-02-23
Linux brew for Travis-CI dependencies


zbeekman:
I am going to look into using [LinuxBrew](http://linuxbrew.sh) to manage Travis-CI dependencies. This may allow us to compile bottled GCC 5.3 and 5.4 in an ubuntu VM and then download and install from the bottle on Travis-CI. Right now using the whitelisted packages and PPAs on ubuntu to get non-ancient gfortran are a real pain.

rouson:
:+1:

zbeekman:
I've revised my thinking on this matter, and I'm not entirely sure Linuxbrew is stable/mature enough to fit the bill here... I think that newish compilers are available with enough frequency through the additional PPAs, and short of that, a docker container would probably be a better alternative if we're going to roll our own solution

PR #177, opened 2016-04-30
Try to ensure g++-5 exists

zbeekman:
install.sh is still failing silently on ubuntu during Travis-CI... Once #150 is merged/completed, this shouldn't be an issue. For the time being I am going to try to ensure that the build progresses a little bit further for GCC, even though I'm anticipating it will still fail (due to a time out)


rouson:
LGTM

PR #180, opened 2016-05-04
Bump default gcc to 6.1

rouson:


zbeekman:
@rouson I haven't had a chance to look at this in detail, but I am concerned about the impact that requiring GCC 6 will have on our test infrastructure before formulae make it into Homebrew and/or whitelisted Ubuntu PPAs.

rouson:
@zbeekman Would it be possible to set up the tests so that they install with an earlier version temporarily?  

./install.sh --with-c=blah --with-c++=blahblah --with-fortran-blahblahblah

If so, I'd be glad to test the above approach manually with 5.3.0 before merging this pull request.

The key concern is that users don't currently have access to the STOP bug fix and also that some users might want to install just for purposes of building 6.1.0.  There's some discussion [online](https://groups.google.com/forum/#!topic/comp.lang.fortran/ZFr1Ymp8l90) at the moment about difficulties with installing 6.1.0 binaries. Our installer could be an alternative.

zbeekman:
Sure default can be bumped so long as minimum is not

rouson:
Great.  I'll test the -with-* flags today with the GCC 5.3 compilers.  Then I'll approve and merge this pull request any required adjustment to the Travis-CI scripts.  I think the solution will take the following form: let the CMake scripts specify the minimum and let install.sh specify the default.  Then it should work to just install with the aforementioned --with-* flags to override the default in a way that meets the minimum.

zbeekman:
Perfect, sounds great!

rouson:
With the additional commits that have been added, this pull request functions properly as follows:
1. If the user specifies compilers with --with-fortran (or -f), --with-c (or -c), and --with-cpp (or -C), then those compilers are used.
2. In the absence of the --with-* arguments, GCC 6.1.0 is the minimum compiler version required to be in PATH or in download/prerequisites in order to avoid triggering the download, build, and installation of GCC 6.1.0.

rouson:
LGTM

PR #181, opened 2016-05-10
Use curl on OS X & append version # to install_path

rouson:


zbeekman:
I never signed off, but pull approve seems to think that I have... not sure why. But it does LGTM!

One note: Homebrew installs to `$(brew --cellar)/pkg-name/pkg.ver.sion/<bin, etc, share, include, etc.>`

zbeekman:
so even though pull approve thinks I have already reviewed this, I'll just be explicit here:

LGTM!

PR #184, opened 2016-05-11
Switch ftp to curl on OS X

rouson:
All OS X downloads now use curl, which is installed by XCode.  Previous uses of wget and ftp have now been switched to curl whenever uname returns "Darwin".


zbeekman:
I don't understand why this failed but previous commits passed... I'm going to re-run this and see what happens

rouson:
LGTM. It appears the tests passed on the second attempt.

zbeekman:
@rouson yes, but at least one of the builds, which used to pass, is now failing silently due to not returning non-zero output status: https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/129755129#L285-L314

rouson:
@zbeekman, ok based your final paragraph above, I assume that the test failure indicates a problem with the Travis-CI setup rather than a problem with install.sh.  I think I"m done with my latest edits.  I'll run one or two more tests and then submit another pull request before we talk or during our call.  I think we're ready to release.

PR #185, opened 2016-05-18
Fix termination when no mpif90

rouson:
This merge corrects several issues with handling the definition of build and install directory names and their creation and detection when previously created.  It also eliminates an error message that occurs when there is no mpif90 present.


zbeekman:
@rouson This looks good and is now passing! Further more, it seems that there is less output now and we may be able to safely remove the hack that was being used to prevent the build from erroring due to excessive output.

rouson:
If there is less output, that's coincidental. I don't think I did anything to reduce the output and would be a little surprised if any change in the output is significant.  I'll create an issue for me to go through all of the installation scripts and replace "echo" and "printf" with a bash3boilerplate output function wherever possible.  Then we can vary the amount of output by setting the LOG_LEVEL.

zbeekman:
I'm going to attempt to adjust the CI scripts to not background the install.sh script by pushing to this branch... if it doesn't work I'll revert the changes, but if it does I'll clean up, merge and release

rouson:
Sounds good.  

D

zbeekman:
It seems linux builds using the install script too! (but OS X had trouble fetching cmake with curl... I suspect it's just a fluke, re-running test now) We should add a flag to run ctest when installing with the install script...

zbeekman:
LGTM

zbeekman:
LGTM

PR #186, opened 2016-05-19
Rev bump

zbeekman:
Prepping for 1.6.0 release

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ set ( CMAKE_BUILD_TYPE "Release"
 set_property ( CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS ${CMAKE_CONFIGURATION_TYPES} )
 
 #Name project and specify source languages
-project(opencoarrays VERSION 1.5.1 LANGUAGES C Fortran)
+project(opencoarrays VERSION 1.6.0 LANGUAGES C Fortran)
 
 #Print an error message on an attempt to build inside the source directory tree:
 if ("${CMAKE_CURRENT_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_BINARY_DIR}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ set ( CMAKE_BUILD_TYPE "Release"
 set_property ( CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS ${CMAKE_CONFIGURATION_TYPES} )
 
 #Name project and specify source languages
-project(opencoarrays VERSION 1.6.0 LANGUAGES C Fortran)
+project(opencoarrays VERSION 1.6.1 LANGUAGES C Fortran)
 
 #Print an error message on an attempt to build inside the source directory tree:
 if ("${CMAKE_CURRENT_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_BINARY_DIR}")


zbeekman:
:shipit:

rouson:
:clap:  Thanks! :D

zbeekman:
@rouson hey if you have a sec can you take a look at my gitter.im message?  I think we should add more stuff to the release notes; doesn't GCC 6.1 combined with OpenCoarrays 1.6 unlock events? Are there other things like this that are enabled from Fortran 2015 or 2008? This might be old news to us, but should probably be advertised and emphasized to users... also updating the build system alone probably doesn't warrant a minor version bump so I think it would be good to include this in the release notes.

rouson:
The short answer is you're right about events. The longer answer is that 5.4 will also include events and I had naively expected that 5.4 would release before 6.1, so although this is chronologically the first release that supports events, it won't be the lowest version number that supports events once 5.4 releases.   Sheesh.... complicated stuff. I'll add something now. I can't think of anything else significant along these lines to add.

zbeekman:
OK, just a quick note about events will excite some users.

szaghi:
I am excited...

rouson:
@szaghi, thanks for letting us know.  We've had more than 850 downloads, but we don't hear from users often.  I'm guessing most people haven't even heard of Fortran 2015 events and there is no other compiler or runtime library of which I'm aware that supports it.  I suspect this is a much bigger deal than most people realize and will ultimately lead to higher-performing applications at a much lower development cost as compared to the alternatives (especially MPI-3 or Fortran 2008 atomics).

zbeekman:
@rouson we're ready to release pending the events update in the release notes... I guess I can take care of that if there's nothing else unless you're doing it... I'm happy to do it if you let me know ASAP, otherwise I'm going to bed and can check for it in the morning.

rouson:
I'm writing it now and should have the language saved in 5 minutes or less.

szaghi:
@rouson totally agree. I am experimenting with FOODIE on Fermi and caf is wonderfull but having opencoarrays on it would be great! I am afraid to ask to Fermi sys admin to install opencoarrays... Cross compiling with gfortran is possible on Fermi (I never used it, I still rest on IBM XL) so I guess that it is possible to install opencoarrays on Fermi. Maybe it is already on it into some nested modules... do you know something about? Or maybe @afanfa ?

rouson:
@szaghi, two responses:
1. Please don't hesitate to ask sys admins to install OpenCoarrays. We would love to add new platforms to our list of [tested platforms](http://www.opencoarrays.org). 
2. You might first try installing OpenCoarrays in your own home directory (or in the appropriate scratch space) on any designed platforms.  We have worked extremely hard over the past 8 months to make the installation process painful.  Our goal is to ensure that at least a functional, default installation results from a user doing nothing more than typing ```./install.sh``` and answering "yes" interactively to all subsequent questions or supplying a stream of "yes" answers non-interactively via "yes | ./install.sh".  If this workflow fails, we will do our best to resolve any related problems submitted via our [Issues page](https://github.com/sourceryinstitute/opencoarrays#issues).

rouson:
Done!  Release 1.6.0 has been posted.  Many thanks, @zbeekman , for your hard work and pushing me to resolve the Travis-CI build test failure.  And thanks,@szaghi, for your feedback and interest.

szaghi:
@rouson Thank you too to exist!

I have just sent an email to a friend of mine working at Cineca (you are in CC in the email): I asked for the correct procedure for installing OpenCoarrays on Fermi. This guy is very smart (an Italian equivalent of Zaak :smile: ), I am confident to obtain good indications.

See you soon.

rouson:
:clap:

PR #190, opened 2016-05-24
Support gcc dev branch builds

rouson:
The following now works to list the available gcc development branches:
```
./install.sh -B gcc
```
and the following will build, for example, the gcc trunk:
```
./install.sh -b trunk -p gcc -i /opt
```


rouson:
LGTM

PR #199, opened 2016-06-06
Non-interactive install options: -y/--yes-to-all

### LINKED ISSUE #191: make install.sh viable in non-interactive mode
Either it is undocumented or unsupported, but I cannot find a way to run `./install.sh` in non-interactive mode, which is critical if I am to use OpenCoarrays in Travis CI.

```sh
opencoarrays will be installed in /home/travis/build/jeffhammond/PRK/PRK-deps/opencoarrays/opencoarrays/1.6.1
Ready to rock and roll? (Y/n)
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

See https://travis-ci.org/jeffhammond/PRK/jobs/133685882 for details.

rouson:
Fixes #191 by adding the following option:
```
./install.sh -y
```
or equivalently
```
./install.sh --yes-to-all
```
will now causes a non-interactive build by assuming affirmative
responses to all user queries.  Exception: if the user passes
an installation prefix (via -i or --install-prefix), the user
will be prompted for a password prior to the "make install" step.
To avoid this, execute as `sudo ./install.sh...` with the
consequence that all steps will execute with sudo privileges.


rouson:
@zbeekman, I recommend adjusting the Travis-CI scripts to use this new option instead of the "yes | ./install.sh" approach.  That would test this new option and would keep our testing more self-contained.

zbeekman:
LGTM

I'll adjust Travis-CI by pushing another commit to this branch before merging

zbeekman:
Ok, Assuming all tests pass, I'm happy to merge this

LGTM

rouson:
Thanks!  I just deleted the branch.

PR #200, opened 2016-06-07
Update Travis-CI to use GCC-6

zbeekman:
If I'm half a clever as I think I am 😋 this should fix broken builds due to GCC upgrades on OS X and let us test GCC 6.1.0 across all OSes.

Note, I have yet to build a bottle for MPICH that uses gcc as the C compiler, so on OS X, clang will be used to compile the OpenCoarrays library, until I can build a bottle... See #65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,12 @@ language: generic
 env:
   global:
     - CACHE="$HOME/.local"
-    - MPICH_VER="3.1.4"
+    - MPICH_VER="3.2"
     - MPICH_URL_HEAD="http://www.mpich.org/static/downloads/$MPICH_VER"
     - MPICH_URL_TAIL="mpich-${MPICH_VER}.tar.gz"
     - MPICH_DIR="$HOME/.local/usr/mpich"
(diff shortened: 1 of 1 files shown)


szaghi:
@zbeekman Sorry to pollute this PR... can you confirm that this travis config successfully uses GNU gcc 6 on Travis Linux Box? The last week I tried without success to switch from 5.x to 6.x trunk... I am surely less than half clever than I think to be :smile:

zbeekman:
> @zbeekman Sorry to pollute this PR... can you confirm that this travis config successfully uses GNU gcc 6 on Travis Linux Box? The last week I tried without success to switch from 5.x to 6.x trunk... I am surely less than half clever than I think to be 😄

I am not installing 6.x trunk, I am installing 6.1.0. It seems to work: https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/135886698#L355-L363

I have some stuff to do to tidy up... noteably, we can't get coverage information with Clang and gcov, so I'm going to build a custom MPICH (and OpenMPI too, probably) bottle for testing on OS X

zbeekman:
:zap: GCC 6.1.0 across the board, as well as MPICH 3.2

PR #206, opened 2016-07-09
Remove legacy files. Cordon off unmaintained files

rouson:
Gfortran produces its own libcaf_single.a to support compilation exclusively for single-image execution.
Rather than to maintain separate source for building the same functionality in OpenCorrays, this commit removes the src/single directory and all references to it in the make and CMake build scripts.

Other changes to src/Makefile:
1. Removed all references to tests. Such references were broken because src/tests has no Makefile.
2. Replaced the previously commented GASNet/ARMCI lines with environemnt-based conditional logic.

The resulting Makefile has no commented executable lines, some of which have been hanging around for a couple of years.  Now all retained lines are either actual comments or actual executable statements, some of which are only enabled when the environment variables GASNET and ARMCI have the value "yes".


rouson:
@afanfa, this pull request contains a massive overhaul of the static Makefile.  Please take a look at this and approve if you're ok with it.

afanfa:
The makefile looks good to me but I would keep libcaf_single without compiling it.

rouson:
After discussion with Alessandro, I'll bring the src/single directory back but will delete the armci directory.

rouson:
LGTM

Issue #202, opened 2016-06-21
Linux Users Unable to Commit


scrasmussen:
The `ends_with_newline()` function in [developer-scripts/git-hooks/pre-commit](https://github.com/sourceryinstitute/opencoarrays/blob/master/developer-scripts/git-hooks/pre-commit) produces an error when being called from a linux machine. 
```
function ends_with_newline() {
    test ! -s "$1" || test $(tail -r -c1 "$1" | xxd -ps) = "0a"
}
```

The error reported is `tail: invalid option -- 'r'`.  The [linux tail](http://linux.die.net/man/1/tail) does not have a `-r` option while the [osx tail](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/tail.1.html) does. The function is only checking the last character so I belief it doesn't matter if the tail output is in reverse order. I think this issue can be fixed by simply removing the `-r` option.

Reproducing the error should be possible by 
1. use linux
2. configuring your local repository with the whitespace settings by running ./developer-scripts/setup-git.sh
3. attempting to commit a file.

zbeekman:
Thanks for the bug report I'll fix this ASAP. In the meantime, you can add `--no-verify` to the failing got command to skip running the hooks.

zbeekman:
Hi @scrasmussen, I just wanted to make sure you saw my previous comment with the work around.

I am working on cleaning up the git hooks; they were cobbled together from some krufty looking code that ships with git and various other examples from around the web. With the help of shellcheck, and more thorough testing I should have some cleaned up versions committed in the not so distant future...

scrasmussen:
Ya, sorry I didn't respond, it worked perfectly. Thanks!

zbeekman:
@scrasmussen The fixed and cleaned up githooks are now available from the latest release and merged into master. Please let me know if you have any further issues... Re-installing them by running the hook-installation script again is probably the fastest way to get them, or just copy them into your local .git/hooks folder.

rouson:
👏

scrasmussen:
@zbeekman I'll let you know if anything doesn't work but so fast so good. Thanks!

PR #209, opened 2016-07-21
Small tweaks

zbeekman:
Various tweaks and bug fixes. Specifically #202 should be fixed and issues with OpenMPI being oversubscribed in the test suite.


zbeekman:
LGTM

PR #213, opened 2016-09-01
Treat user-specified install prefix as full path

rouson:
previously subdirectories named after the package and version were
appended to user-specified install prefixes so

./install.sh -p gcc -i /opt -I 6.1.0

would produce an installation directory named "/opt/gcc/6.1.0".
This differs from how most packages interpret the install prefix
and can result in redundancy if the user has already included such
information in the -i argument. With this commmit, the bare user-
-specified install prefix will be used with nothing appended.  If
no -i argument is present, the default still contains the package
name and version: ${OPENCOARRAYS_SRC_DIR}/prerequisites/installations/gcc/6.1.0)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -194,9 +194,6 @@ info  "-y (--yes-to-all):       ${arg_y}"
 this_script="$(basename "$0")"
 export this_script
 
-export install_path="${arg_i%/}"
-info "install_path=\"${install_path}\""
-
 export num_threads="${arg_j}"
 info "num_threads=\"${arg_j}\""
 
--- a/install.sh
+++ b/install.sh
@@ -194,6 +194,9 @@ info  "-y (--yes-to-all):       ${arg_y}"
 this_script="$(basename "$0")"
 export this_script
 
+export install_path="${arg_i%/}"
+info "install_path=\"${install_path}\""
+
 export num_threads="${arg_j}"
 info "num_threads=\"${arg_j}\""
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ env:
     - MPICH_URL_HEAD="http://www.mpich.org/static/downloads/$MPICH_VER"
     - MPICH_URL_TAIL="mpich-${MPICH_VER}.tar.gz"
     - MPICH_DIR="$HOME/.local/usr/mpich"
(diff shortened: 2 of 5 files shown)


install.sh
```diff
@@ -194,9 +194,6 @@ info  "-y (--yes-to-all):       ${arg_y}"
 this_script="$(basename "$0")"
 export this_script
 
-export install_path="${arg_i%/}"
-info "install_path=\"${install_path}\""
-
```
zbeekman: It seems that there is a problem here: Later, on line [296](https://github.com/sourceryinstitute/opencoarrays/pull/213/files#diff-3fbb47e318cd8802bd325e7da9aaabe8R296) `install_path` is de-referenced before being defined. It's unclear to me why this [only registers as an error on Linux](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/156959330#L426) but, even though [the OS X script complains about this](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/156959327#L222), it [doesn't register as an error with Travis-CI](https://travis-ci.org/sourceryinstitute/opencoarrays/builds/156959324)

zbeekman:
I just pushed up changes to this branch that should resolve the Travis-CI/MPICH related issues. It seems that there was an issue with MPICH that was causing some people to experience segfaults. This issue has been fixed upstream, and Homebrew released a patch for the MPICH formula, which is what invalidated our custom MPICH bottle. I don't think the upstream MPICH bug effected us at all, but I'm not 100% about that...

Any way, we still need to fix the de-referencing of undefined vars here to be able to merge these changes.

zbeekman:
I can confirm that my changes have fixed the outdated MPICH bottle issue on Travis-CI. We just need to fix the undefined references now. I probably won't have any more time to look at this today.

install.sh
rouson: I'll just re-insert the lines.  I'd like to get this into a release ASAP and then point people on comp.lang.fortran to the release in case anyone wishes to install the current gfortran 7.0.0 trunk to try out user-defined derived type input/output (UD-DTIO), which two gfortran developers just committed to the trunk based in part on promised funding from Sourcery Institute.

rouson:
LGTM

PR #215, opened 2016-09-06
Enable bare install-prefix for OpenCoarrays itself

rouson:
This change was already made for installation of other packages.
Now

./install.sh --install-prefix /foo/bar/

will result in the installation of OpenCoarrays to /foo/bar without
appending "opencoarrays/<version-number>" to the installation path.


zbeekman:
LGTM

PR #218, opened 2016-09-20
Fixes #216

### LINKED ISSUE #216: Installer fails to pick up some long-form arguments
@zbeekman, the installer fails to detect the "--branch trunk" argument in the following:
```
./install.sh -p gcc --branch trunk
```
Presumably the problem is in [parse_command_line.sh](https://github.com/sourceryinstitute/opencoarrays/blob/master/prerequisites/use-case/parse_command_line.sh).

rouson:
It appears the problem was a mismatch between an argument name in
install.sh-usage (--branch) and the corresponding argument in
prerequisites/build.sh (--install-branch).


zbeekman:
LGTM

PR #219, opened 2016-09-22
Fixes to support gcc-7's caf-interface.

vehre:


zbeekman:
@vehre can you please [sign the CLA](https://cla-assistant.io/sourceryinstitute/opencoarrays?pullRequest=219)? Thanks!

zbeekman:
@vehre @rouson Do these changes *require* GCC-7/trunk to run correctly? i.e. are they backwards compatible with GCC-5 and GCC-6?

vehre:
These fixes do need gcc-7 with a patch that is still in testing. That is also the reason why I have named the branch gcc-7.

zbeekman:
> These fixes do need gcc-7 with a patch that is still in testing. That is also the reason why I have named the branch gcc-7.

@vehre Interesting. Right now the PullRequest has been opened against `master`. (Your local branch is `gcc-7` but this PR is attempting to merge into `master` which is a stable branch.) All of our automated testing infrastructure assumes GCC < 7, so the code is compiling successfully against gcc-6 but [encounters severe runtime errors during testing with gcc-6](https://travis-ci.org/sourceryinstitute/opencoarrays/builds/161923112).

Would it be possible to wrap GCC-7 dependent stuff in `#ifdef`s and  default to gcc 5 & 6 compatible code? It should be pretty straight forwards to detect if GCC >= 7 in the config logic and add the correct `#define`s.

vehre:
Well, I don't have write privileges on the opencoarrays repo, therefore I could only create the gcc-7 branch on my private copy. I based it on opencoarray's master. Therefore the relationship.

ifdef'fing the stuff might be good idea. It is only needed for the caf_register though, because the caf_*_by_ref functions are new and not generated by any pre-gcc7. For size matters they should be ifdeffed, too.

vehre:
It can't run with gcc-6. The interface of the caf_register routine has changed and this will lead the program to errorneously write into alien memory. It should crash instead of only take more time.
Furthermore, are tests 10 and 24 still failing with patch. I cite the mail I send to Damian about it:
Dear Damian,

I have send you a pull request to add the required changes for gcc-7's
caf-interface. This also includes a minimalistic caf_get_by_ref to caf_get
adapter. Minimalistic in the sense, that it adapts most of what it is able to.
With these changes and the update of gcc-trunk I have only three testcases
failing:

	  1 - initialize_mpi (Failed)      -- I have no clue, why this fails.
The reason for failure here is not obvious to me and I don't think it is caused
by the changes to the caf-interface.

         10 - whole_get_array (Failed)
This can be fixed, when allocatable arrays on the right-hand side of a coarray
get operation are not unnecessarily declared allocatable although they get
allocated just once with a fixed size and never change.

        24 - coarray_navier_stokes (Failed)
The same as with 10.


For testcase 22 - coarray_burgers_pde I had to change the number of process to
4 in the makefile to get it running reliably with the same result. With 2
processes it always asserts.

At the moment I am preparing the patch for gcc-7 and doing regression tests. It
will hit trunk today I think.

Regards,
	Andre

The patch for gcc-7-trunk is published on the mailing list a few second ago. Just awaiting review.

zbeekman:
> ifdef'fing the stuff might be good idea. It is only needed for the caf_register though, because the caf_*_by_ref functions are new and not generated by any pre-gcc7. For size matters they should be ifdeffed, too.

I'm tentatively going to talk to Damian later today (2pm Eastern US), but I'm guessing that this will be the best approach. How long would it take to implement? If it's easy, go ahead and push another commit to your private gcc-7 branch with the conditional inclusions. That way OpenCoarrays will still be backwards compatible.

vehre:
I expect the ifdef guards to be present by tomorrow. May be already today. Just going to take a jog now and wait where Damian is ok with what I did so far.

rouson:
@vehre, please let me know as soon as the most recent GCC patch you mentioned has been applied to the trunk.  I succeeded in completing a fresh clone of the gcc-7 trunk today.  Using today's gcc-7 and the gcc-7 branch of the vehre/opencoarrays fork, I can build OpenCoarrays, but all tests fail.  I assume that your latest patch will be required to get to the state wherein only tests 1, 10, and 24 fail as you mentioned above.

@afanfa, please take a look at the above three tests if at your earliest convenience.  My first lecture will be on Monday and I'll be handing out the virtual machine to the students then.  For now, I'll give them a version that uses the gcc-7 trunk from 2016-09-04, which means I won't have access to the UD-DTIO feature that Paul developed under contract for Sourcery Institute. I hope to be able to use a newer gcc-7 trunk soon, but will wait until all tests are passing.

rouson:
LGTM

@zbeekman and @afanfa, because this pull request is against a branch (gcc-7) and passes all test when built against GCC 6.1.0, it seems safe for me to merge it.  I'll do so.  Then, one of us should create a pull request for the gcc-7 branch and we should jointly decide about merging this into the master branch.

PR #222, opened 2016-09-29
Fix for sync images issue and some cleanup

vehre:
Hi all,

this pull request adds the improved sync images routine we have talked about. The routine functions also under heavy load and has less race conditions.

Note, testcase 20 hello_multiverse still fails with gcc-7, because of a defect in gcc-7's string building. Try   

character(len=MAX_STRING) :: greeting ! no coarray here
write(greeting,"(2(a,i2))") "Greetings from image ",this_image()," of ",num_images()
print *,greeting
end

and marvel.

Note 2: Tests only succeed with trunk with my patch at: 

https://gcc.gnu.org/ml/fortran/2016-09/msg00188.html

I intent to notify you ones the patch has been approved.

This completes the work I intent to do on this specific issue on opencoarrays.

Regards,
Andre


rouson:
LGTM

rouson:
I also confirmed today that this branch builds and passes all tests with GCC 5.4.0 on Linux and MacOS.

zbeekman:
@rouson  This PR was against master. Did you intend to merge into master? I thought we were keeping this on the gcc-7 branch. Since it doesn't break backwards compatibility, I think it's probably fine either way.

rouson:
@zbeekman, thanks for taking a look at this.  As I noted above, this branch passes all tests with GCC 5 and 6 so it doesn't break backwards compatibility unless you're referring to GCC 7, in which case I think it's ok to break backwards compatibility with a pre-release version.  I'll await your concurrence before merging.

zbeekman:
Hi @rouson, you already merged this PR into master. It's fine by me, but it makes our gcc-7 branch redundant

PR #223, opened 2016-09-30
Temporarily disallow GCC 7 unless explicitly requested.

rouson:
Merge Andre's gcc-7 branch with build script fix to prevent gcc-7 builds, for the time being

> Hi all,
> 
> this pull request adds the improved sync images routine we have talked about. The routine functions also under heavy load and has less race conditions.
> 
> Note, testcase 20 hello_multiverse still fails with gcc-7, because of a defect in gcc-7's string building. Try
> 
> character(len=MAX_STRING) :: greeting ! no coarray here
> write(greeting,"(2(a,i2))") "Greetings from image ",this_image()," of ",num_images()
> print *,greeting
> end
> 
> and marvel.
> 
> Note 2: Tests only succeed with trunk with my patch at:
> 
> https://gcc.gnu.org/ml/fortran/2016-09/msg00188.html
> 
> I intent to notify you ones the patch has been approved.
> 
> This completes the work I intent to do on this specific issue on opencoarrays.
> 
> Regards,
> Andre


zbeekman:
LGTM, but lets still wait until the tests finish for extra assurance.

zbeekman:
@vehre In https://github.com/sourceryinstitute/opencoarrays/pull/223/commits/3ddd291c8852cdcea3bf9d6aec753dd57b974bcb you said:

> Still failing tests:
> hello_multiverse: This test fails in general with the gcc-7 fortran compiler,
> 	because the string is not composed correctly. This is independent
> 	of coarrays being present or not.

Which string are you referring to? Is it this one: https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/integration/coarrayHelloWorld/hello_multiverse.F90#L45

If so, what's wrong with how it is composed?

src/tests/unit/send-get/get_with_offset_1d.f90
```diff
@@ -27,5 +27,5 @@ program get_offset_1d
     enddo
     write(*,*) 'Test passed.'
   endif
-
+sync all
```
zbeekman: @afanfa can you please look at this?

@vehre can you please elaborate more on why this is necessary?

PR #224, opened 2016-09-30
Fix for sync images issue and some cleanup (#222)

rouson:
* Fixes to support gcc-7's caf-interface.

* Added preproccessor guards.

* First steps to implemeting full featured caf_get_by_ref().
WARNING! Errorneous at the moment.

* Reverted change for burgers test in CMakeLists.txt.
Fixed most issues. One still needs a patch to gfortran. Therefore
the pde-solver might fail with an unpatched gfortran still.

* Removed define sticking to gcc-7.

* Fixed typo.

* Send-get/get_with_offset_1d.f90: Needs sync all at the end of the main program
	or the scope where array descriptors are stored may be already
	destroyed when accessed by other (slower) images.
mpi_caf.c: Fixing sync_images, when a stopped image is slow to reach its
	caf_finalize and set the stopped status, while other images have read an
	ok status already and now wait on its send. Realized by using waitany
	and propagating a stopped image status.
	Small speed up in get_by_ref and smaller memory footprint.

Still failing tests:
hello_multiverse: This test fails in general with the gcc-7 fortran compiler,
	because the string is not composed correctly. This is independent
	of coarrays being present or not.

### DIFF
--- a/install.sh-usage
+++ b/install.sh-usage
@@ -1,4 +1,4 @@
-  -b --install-branch [arg]           Install the specified repository development branch.
+  -b --install-branch [arg]   Install the specified repository development branch.
   -B --list-branches [arg]    List the available branches in the specified package's repository.
   -c --with-c [arg]           Use specified C compiler. 
   -C --with-cxx [arg]         Use specified C++ compiler. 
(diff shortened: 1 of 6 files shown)


rouson:
LGTM

PR #226, opened 2016-10-02
Gcc 7

zbeekman:
Redo merge into OpenCoarrays main repo without squashed merge

> Hi all,
> 
> this pull request adds the improved sync images routine we have talked about. The routine functions also under heavy load and has less race conditions.
> 
> Note, testcase 20 hello_multiverse still fails with gcc-7, because of a defect in gcc-7's string building. Try
> 
> character(len=MAX_STRING) :: greeting ! no coarray here
> write(greeting,"(2(a,i2))") "Greetings from image ",this_image()," of ",num_images()
> print *,greeting
> end
> 
> and marvel.
> 
> Note 2: Tests only succeed with trunk with my patch at:
> 
> https://gcc.gnu.org/ml/fortran/2016-09/msg00188.html
> 
> I intent to notify you ones the patch has been approved.
> 
> This completes the work I intent to do on this specific issue on opencoarrays.
> 
> Regards,
> Andre


zbeekman:
Superseded by pull request #223

PR #228, opened 2016-10-14
Fix GCC download_prerequisites script

rouson:
The installation system now modifies the download_prerequisites script that GCC provides to replaced the hardwired invocation of wget get with whichever download mechanism the OpenCoarrays build system has chosen.  This is most important on MacOS, which ships with curl but not wget.  It's also helpful if a system doesn't have curl or wget, in which case ftp is used.


rouson:
LGTM

PR #230, opened 2016-10-15
Fixes #229

### LINKED ISSUE #229: Installation fails if gfortran version < 4.9 is in PATH
If gfortran exists in the users's PATH, then the OpenCoarrays installation scripts invoke gfortran [here](https://github.com/sourceryinstitute/opencoarrays/blob/master/prerequisites/install-functions/find_or_install.sh#L167) or [here](https://github.com/sourceryinstitute/opencoarrays/blob/master/prerequisites/install-functions/find_or_install.sh#L264) to compile a simple program that prints the compiler version using the Fortran 2003 "compiler_version" function.  Older versions of gfortran do not support this function so compilation fails and the installation script dies.  Probably a few applications of "|| true" would make this more robust so that the installation simply treats an inability to compile the test program as an indication of an insufficiently recent version of gfortran.

rouson:
The installation no longer fails when a GCC installation is in the
PATH and is too old to support the Fortran 2003 compiler_version
function invocation in prerequisites/acceptable_compiler.f90.

### DIFF
--- a/prerequisites/install-functions/find_or_install.sh
+++ b/prerequisites/install-functions/find_or_install.sh
@@ -163,14 +163,17 @@ find_or_install()
 
         else
 
-          echo -e "$this_script: Checking whether $executable in PATH wraps gfortran version >= $(./build.sh -V gcc) and < 7.0.0 ... "
-          $executable acceptable_compiler.f90 -o acceptable_compiler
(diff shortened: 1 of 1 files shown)


rouson:
LGTM

Issue #225, opened 2016-09-30
Events tests


rouson:
@afanfa, I don't believe we have any tests for Fortran 2015 events in our test suite.  I'll add some.  If you have any you'd like to add please do so.

PR #231, opened 2016-10-17
Add Fortran 2015 events test

rouson:


rouson:
Fixes #225

zbeekman:
@rouson I think the tests failed. Did you merge this intentionally?

rouson:
@zbeekman. Yes, I made some changes on fix-gcc-download-prerequisites-script that should have been on add-events-tests.  Merging this just puts the commits onto the intended branch, add-events-tests, which will not be merged into master until all tests pass.

I'm seeing two strange test failures locally and am attempting to verify whether the Travis-CI builds encounter the same errors, which is one reason I'm pushing code that fails locally. It appears there might have been a regression that affects both gfortran 5.4.0 and 6.2.0. We can talk about this tomorrow.  Hopefully, @afanfa and @vehre can join the weekly call and shed some light.  In both cases, the failure stems from internal file output [here](https://github.com/sourceryinstitute/opencoarrays/blob/add-events-tests/src/tests/integration/coarrayHelloWorld/hello_multiverse.F90#L45) and [here](https://github.com/sourceryinstitute/opencoarrays/blob/add-events-tests/src/tests/integration/events/async-hello.f90#L32).  Both generate the following error message:

Internal Error: get_unit(): Bad internal unit KIND

Given that this is an internal error, it is by definition a compiler bug.  Two gfortran developers recently made commits to the gfortran trunk related to internal file I/O, but I wasn't aware that any of their commits had been backported to 5.4.0 or 6.2.0 so this is really strange.    And I can't build OpenCoarrays with the latest 7.0.0 trunk because of the situation with the allocatable components interface mismatches. 

Everything works fine when I build the aforementioned two tests manually with a checkout of the trunk dated 20160904. That appears to have been a golden era when everything I need for my current class worked.  I'm so glad I saved that snapshot. :O

vehre:
Why can't you build gcc-7-trunk? What interface mismatches?

rouson:
I set up our installation script to disallow building with GCC 7 because of the differences between the OpenCoarrays calls in the trunk and the actual OpenCoarrays implementation that were introduced last month.  I haven't attempted to rebuild since then. Is it now possible to build OpenCoarrays with the trunk?  If so, how many of the OpenCoarrays tests fail?

zbeekman:
Ah I see. I missed that this wasn't master on my phone

PR #232, opened 2016-10-17
Add events tests

### LINKED ISSUE #225: Events tests

rouson:
Fixes #225 

This pull request adds an integration test of the asynchronous events features of the [draft Fortran 2015 standard](http://isotc.iso.org/livelink/livelink?func=ll&objId=17852450&objAction=Open).  The test is an integrated use of the following features:
* The event_type derived type.
* The event_post, event_query, and event_wait subroutines.
The  top-level CMakeLists.txt file exercises the above features on 2 images.  The test source is written to work on any number of images, but it only tests the features on at most 9 images.


rouson:
@afanfa, please take a look at the new test and think about how it might be made more feature-complete.  For example, if the event subroutine calls accept additional arguments that are not tested, please add those arguments if it's easy to do so.  Also, if there is any way to write more tightly focused, shorter unit tests for each of the subroutines, please add those in [src/tests/unit](https://github.com/sourceryinstitute/opencoarrays/tree/master/src/tests/unit).

rouson:
@zbeekman, let's include this pull request in the upcoming release if the tests pass and it's not to late to include this.

rouson:
@zbeekman, if I'm understanding the output correctly, then my latest push corrected any problems with this pull request and all tests are now passing on Travis-CI.  I've received word from a gfortran developer that the problem might relate to compiling with older versions of gfortran but linking with newer versions of libgfortran. I'm not sure why this would be happening on my system, but I suppose MacPorts could be to blame.  I imagine it's ok to merge this and release.

zbeekman:
Let me double check on my system. Either way it sounds like a gfortran
problem not an opencoarrays problem but it would be nice not to alarm users
with failing tests.

rouson:
LGTM

zbeekman:
LGTM (once tests finish)

PR #237, opened 2016-11-01
Pull request on wrong preprocessor symbol for HAVE_INT128

vehre:
This pull-request fixes the issue the on certain systems the HAVE_INT_128T guard was available although the int128_t was not. The commit syncs the guard to be HAVE_GFC_INTEGER_16 and the type to be __int128 like it is used in previous code.


zbeekman:
@vehre I'm confused as to why github wants to merge all of these commits into master: I thought we had already merged them, albeit perhaps onto a different branch. I think I'm going to just cherry pick 	32fc2e4 onto a new branch, and open that as the pull request if you don't mind.

zbeekman:
Superseded by #250

PR #241, opened 2016-11-03
Cleanup tests

### LINKED ISSUE #239: Remove non-standard abort() calls in tests
With the aim of making OpenCoarrays viable for use with multiple compilers, please ensure that all code that hits the master branch is standard-conforming.  An example of non-standard code is the "call abort()" appearing throughout  the test suite, including but definitely not limited to the following locations:

1. [Line 26](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/send-get/get_with_offset_1d.f90#L26) of  get_with_offset_1d.f90. 
2. [Line 45](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/send-get/get_array_test.f90#L45) and other lines in get_array_test.f90.
3. [Line 173](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/send-get/old_get_array_test.f90#L173) and other lines in old_get_array_test.f90.
4. [Line 133](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/send-get/sameloc.f90#L133) of sameloc.f90.

and many other locations.  Presumably these only terminate a single image and should therefore be replaced with ```stop```.  Alternatively, if global termination is intended, they should be replaced with ```error stop```.   

If there is a compelling reason to include non-standard code in the repository, let's document it. I will submit a separate issue to help address the state of documentation in the tests.

zbeekman:
Cleanup and enable more tests

A logic error exists in one of the tests, therefore it has not been enabled, but all others have.

Also remove non-standard calls to `abort()` and replace with `error stop` (Fixes #239)


rouson:
@zbeekman the GCC [documentation](https://gcc.gnu.org/onlinedocs/gcc-4.3.6/gfortran/ABORT.html) for ```abort()``` states that it "causes immediate termination of the program" but makes no mention of images.  I would have guessed that it behaves like ```stop``` and therefore halts only one image, but some simple testing of it suggests it behaves like ```error stop``` and halts all images.  It would be great if @afanfa could confirm the intended behavior.

zbeekman:
I reviewed the tests and the intent in each case seems to be to error out
and halt all images.

zbeekman:
> 10% coverage increase :-)

LGTM

zbeekman:
Confirmed that no new test failures are created by this when using GCC-7 (trunk)

PR #248, opened 2016-11-10
Support building shared libs

### LINKED ISSUE #245: RFE: Support building shared libraries
Attempting to build shared libraries results in:
```
CMake Error at src/mpi/CMakeLists.txt:27 (install):
  install TARGETS given no LIBRARY DESTINATION for shared library target
  "caf_mpi".
```
Is there a specific reason that this isn't supported?

zbeekman:
Fixes #245

### DIFF
--- a/src/mpi/CMakeLists.txt
+++ b/src/mpi/CMakeLists.txt
@@ -17,6 +17,13 @@ endif()
 
 add_library(caf_mpi mpi_caf.c ../common/caf_auxiliary.c ../extensions/opencoarrays.F90)
 
+set_target_properties ( caf_mpi
+  PROPERTIES
+  SOVERSION ${PROJECT_VERSION.MAJOR}.${PROJECT_VERSION.MINOR}
+  VERSION ${PROJECT_VERSION}
+  )
+
+
 target_include_directories(caf_mpi PRIVATE ${MPI_C_INCLUDE_PATH})
 if (gfortran_compiler)
   target_compile_options(caf_mpi INTERFACE -fcoarray=lib)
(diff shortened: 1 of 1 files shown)


zbeekman:
LGTM

@apoplawski if you have a chance to check this out and ensure it suits your needs, please do so. Cheers!

zbeekman:
LGTM

opoplawski:
You almost certainly do not want to set the SOVERSION to be based on the project version, but instead start it at 0 and increment it only when breaking ABI compatibility.

zbeekman:
@opoplawski: ah here is the comment I was looking for in #248 

Can you please elaborate on why we should start at 0?

Btw, CMake has infrastructure for managing some of this by creating the appropriate symbolic links (`libcaf_mpi.so ==> libcaf_mpi.so.MAJOR ==> libcaf_mpi.so.MAJOR.MINOR) and has mechanisms for controlling compatibility (via `find_package(opencoarrays)`)...

We are using semantic versioning, which, in general, is a necessary but *not* sufficient condition for ABI compatibility---I agree with you there.

Perhaps I am missing other cases for when ABI compatibility can break despite not bumping the semver major version, but the only other one I can think of is due to the version of the compiler being used. The ABI  compatibility is dependent on the version of GCC the library was built with, since it interfaces with GCC internals and the gfortran runtime. The library itself is written completely in C, so I don't believe that other Fortran ABI issues are relevant. Therefore, ABI compatibility issues could arise if the user tries to link with OpenCoarrays using a different version of GCC than the library was built with. But this isn't changes in OpenCoarrays resulting in differences in the ABI, it's changes in the build process.

Can you envision any other cases that would break the ABI without having to bump the semver major version?

PR #251, opened 2016-11-11
Expose caf_init and caf_finalize in module to allow use with f2py, see #204

### LINKED ISSUE #204: Expose *_caf_init(), *caf_finalize()
Hi,

I've been experimenting with running CAF under Python using f2py.  Its a bit hacky at the moment (proof of concept time) but I can now compile and run CAF programs under Python3 using f2py3.5, OpenCoarrays 1.6.0 and OpenMPI 1.10.3 on Debian.

So I have a Fortran module:

```
module caftest
contains
  subroutine bar
    implicit none
    integer:: np, me

    np = num_images()
    me = this_image()
    write (*,*) "I am image ", me, " of ", np
  end subroutine bar
end module
```

caftest.py:

```
#!/usr/bin/env python3
# hacky setup code here
caftest.bar()
```

mpiexec -n 2 ./caftest.py
 I am image            1  of            2
 I am image            2  of            2

To do this, I've needed to call _gfortran_caf_init() and _gfortran_caf_finalize() from the Python / f2py layer, in order to setup mpi_init, etc.  But there is no coarray.h, API  exposed, so this was a little .. unorthodox. Can they be exposed?

Regards
Alastair McKinstry

zbeekman:
Fixes #204

 At configure time, choose whether to expose the `caf_init` and
 `caf_finalize` functions in the opencoarrays (extension) module by using `-DCAF_EXPOSE_INIT_FINALIZE`.

N.B. the semantics of using C pointers with Fortran still mystifies me a bit, and I don't really understand what the point of passing `argc` and `argv` to `MPI_init` is (since I always use fortran, where the only arg is `ierr`). So for my interface to `caf_init` which takes `argc` and `argv` as arguments, I have declared both as `type(c_ptr)` so it is the users responsibility to pass the appropriate `c_ptr` in to them. In my testing I just passed `c_null_ptr`s and it seemed to work fine. Also, please note, that the `caf` compiler wrapper script does not pass in the installed module file for open coarrays aware (OCA) compilers (GCC/GFortran) to the compiler, so you'll have to add something like `-J/usr/local/mod` or whatever to the `caf` script if you use it and want the compiler to find your installed module file and compile time.

@amckinstry: will this suit your needs? Please take a look and let me know if you need any changes.

### DIFF
--- a/src/extensions/opencoarrays.F90
+++ b/src/extensions/opencoarrays.F90
@@ -51,6 +51,10 @@ module opencoarrays
     integer(atomic_int_kind), allocatable :: atom[:]
   end type
 #endif
+#ifdef EXPOSE_INIT_FINALIZE
+  public :: caf_init
+  public :: caf_finalize
+#endif
 
   ! Generic interface to co_broadcast with implementations for various types, kinds, and ranks
   interface co_reduce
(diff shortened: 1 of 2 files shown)


zbeekman:
LGTM

Issue #234, opened 2016-10-20
GCC download_prerequisites script edit fails for branch builds


rouson:
In the latest release of OpenCoarrays, the installer edits GCC's download_prerequisites script if gfortran is built as part of the OpenCoarrays installation.  The edit switches the script's download mechanism from wget to whichever one the script has chosen for the given environment (alternatives include curl and ftp).  When using OpenCoarrays to install an unreleased branch of GCC, wget gets switch to svn, which doesn't work for downloading the release tar balls of the GCC prerequisites.  In such cases, the script needs to use a different download mechanism for the GCC prerequisites from the one it's using for checking out the GCC branch.

PR #252, opened 2016-11-12
Fixes issue #234

rouson:


zbeekman:
@rouson this has gotten out of sync with master. We should update either via the "update branch" button in this github UI or via a pull/merge or rebase on the command line (rebase will require a forced push though... so maybe just a pull/merge). I also wanted to merge #204 soon, but was planning to wait the 24 hours or for an explicit code review. It may make sense to merge that first (since the tests all finished) and then merge update this, but I won't do so unless I get an explicit 👍 or 24 hours pass.

rouson:
I just clicked the "Update branch" button on the GitHub web interface.

zbeekman:
I just tested this on the SI Lubuntu image where it had been failing for me before. GCC-trunk is still building, but it has progressed far enough to see that the changes are behaving correctly, let's

:ship: it!

Issue #79, opened 2015-12-19
Review installation logic, files and directories


zbeekman:
- Add an `uninstall` target so users can easily remove OCA (now #89)
- Investigate which files are being installed, to where, and why
  - right now it seems that some of the tests are being installed; these should probably only be run at build time `cmake .. && make && make test && make install` is the canonical order of operations... no need to install tests.
- Fortran and .mod files complicate this a little bit, when on systems with multiple fortran compilers. I developed what I think is a very good system for https://github.com/jacobwilliams/JSON-Fortran

@zbeekman I'm adding an item to this list:
- Write unit tests for the installation scripts.  

I started this with a [commit](https://github.com/sourceryinstitute/opencoarrays/commit/91f8fd1cc1ac3373b1ab6a46d91aad6c25c2d492) pushed last night.  The motivation was that my work on inserting `find_or_install gcc` near the beginning of [build_opencoarrays.sh](https://github.com/sourceryinstitute/opencoarrays/blob/master/prerequisites/install-functions/build_opencoarrays.sh) necessitates changing some stack management in [find_or_install.sh](https://github.com/sourceryinstitute/opencoarrays/blob/master/prerequisites/install-functions/find_or_install.sh) and the stack function I'll most likely use is one that I wrote, but hadn't sufficiently tested. The new tests have already identified some issues in stack.sh and helped define the boundaries of what is safe to use.

rouson:
* Handle additional missing prerequisites: wget, svn, and m4. 
  * When wget and/or svn is missing, print a URL for each file that needs to be downloaded and direct the user download the file via a web browser.
  * When m4 is missing, attempt to install it.

zbeekman:
Just curious, has missing wget, and svn been handled yet? I am expanding the test matrix to cover the install script.

rouson:
@zbeekman, I'm not sure what you mean but yes, that's the subject of my most recent commit merged into master.

zbeekman:
Ah yes, I thought I remembered that this was added, but perhaps I need to pull in the latest commits from master to my branch. Thanks for clarifying.

zbeekman:
@rouson does d5783167707dfb8d2dc973414fa14cd483c57a20 handle both svn/subversion *and* wget? (I looked it over but it's quite sophisticated, and will take some time to fully digest)

rouson:
@zbeekman, short answer: yes.  

[...]

Issue #39, opened 2015-12-04
Better CMake compiler/MPI passing


zbeekman:
I'm wondering how common it is to pass the MPI wrappers to CMake directly. The folks @homebrew seem to be under the impression that this is somewhat abnormal. I'm going to inquire on the CMake list and see what the consensus is there.

zbeekman:
Yes, it seems that the findMPI module might be the way to go. At least one person thinks that using the wrappers from CMake is a bad idea: https://cmake.org/pipermail/cmake/2011-June/045037.html

Also see homebrew/homebrew#46547

zbeekman:
Here is some documentation for using the `FindMPI` CMake module: https://cmake.org/cmake/help/v3.4/module/FindMPI.html

zbeekman:
Here is the documentation for `FindMPI`: https://cmake.org/cmake/help/v3.4/module/FindMPI.html

This should probably be used along with CMake's `configure_file` capabilities to substitute in the right EXE name, flags etc. in the CAF wrapper scripts.

zbeekman:
OK, 193546d seems to solve this quite elegantly, if I do say so myself. I would like to rework the `caf` and `caf_run` scripts to use `configure_file` and full paths to the compiler and MPI implementation, and also run it through shellcheck. I'll open a new issue for this, I think.

Issue #22, opened 2015-11-23
Homebrew Formula?


zbeekman:
I wanted to gauge interest/support for creating and submitting a Homebrew Formula to build & install opencoarrays.
- Pros:
  1. Better package manager for OS X (IMO)
  2. [Reach a larger user base, including those who shun MacPorts](https://www.google.com/trends/explore#cmpt=q&q=Homebrew,+MacPorts,+fink&cat=0-5-32)
  3. pre-requisites (CMake, MPICH, GCC/GFortran---5.2) managed and installed via HomeBrew, and supports CMake build & install systems natively.
- Cons:
  1. No download tracking feature, that I am aware of....

sourceryinstitute:
I'd like for OpenCoarrays to be available by any and every mechanism possible. Are you offering to assist with the set-up or asking about our plans.  Help is always welcome.  We haven't scoped it out, but Alessandro and I are talking tomorrow and will add this to the agenda. Please create an entry for this in our Issues page. 

D

zbeekman:
I'm offering to take a crack at doing it myself. I am the maintainer of a number of Fortran related Homebrew Formulae: JSON-Fortran, FoBiS.py, and FORD.

zbeekman:
It looks like there may be a way to track some of the usage using Homebrew: https://github.com/Homebrew/homebrew/issues/46289#issuecomment-159009961

(A homebrew "bottle" is a package of binaries which is built when the formula is tested. Unless the user requests an install with non-standard options or from the "HEAD" source, should one be provided in the formula, the binary assets built during the Formula testing are downloaded and installed, rather than building the assets from source again. A bottle is created for each of the three most recent OS X releases, and Homebrew fetches and installs the appropriate one.)

rouson:
> On Nov 23, 2015, at 9:50 AM, Izaak Beekman <notifications@github.com> wrote:
> 
> I'm offering to take a crack at doing it myself. I am the maintainer of a number of Fortran related Homebrew Formulae: JSON-Fortran, FoBiS.py, and FORD.
> 
> 

Awesome!  Thanks tons.  We look forward to hearing how it goes.

D

zbeekman:
If it's accepted, then the statistics tracking page will look something like this: https://bintray.com/homebrew/bottles/a2ps/view#statistics

If I'm successful and it is accepted, I would encourage you to drive people towards Homebrew over MacPorts for better download tracking and ease of use... but that's up to you guys.

[...]

Issue #246, opened 2016-11-08
Add installer detection of gfortran/mpich mismatch


rouson:
A common installation failure mode arises when OpenCoarrays is built with an MPICH installation that itself was built by different gfortran version than the gfortran version currently invoked via `mpif90`.  Detecting this failure mode in the OpenCoarrays installer has long been on my To Do list.  This issue is a reminder to work on it.  Detecting and preventing such subtle and hard-to-diagnose problems is one of the main aims of the OpenCoarrays installer.

zbeekman:
There may be some cases when this isn't a problem, at least once we've decoupled the need to explicitly pass in `FC=mpif90` to CMake and rely more heavily on `find_package()`.

One big issue stems from the `.mod` files and their formats, which is non-standardized and radically different between compiler vendors, and sometimes changes from minor version to minor version within a compiler. (Gfortran has a reputation for this...) However, just because the `.mod` file format changed doesn't mean that the version of MPI previously built is incompatible with the current OpenCoarrays and Gfortran toolchain: the ABI of the object files may still be the same.

Furthermore, the only hard dependency of the actual library is on the C MPI interface: if my understanding is correct, we only depend on the Fortran MPI interface for the [OpenCoarrays wrapper module](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/extensions/opencoarrays.F90) and perhaps some tests or system introspection.

I think the best option for detecting this would be through system introspection at configure/cmake time. We can try to build and run an MPI hello world Fortran program. If we're using a recent GCC then we could even consider skipping the OpenCoarrays wrapper module and any tests in which `use mpi` makes an appearance. Furthermore, since it's possible that the ABI for object files etc. is consistent, even if the ABI for `.mod` files isn't, we could also try using `include mpi.h` instead of `use mpi`

rouson:
Thanks for writing this thorough and thoughtful response.  It helped me remember the details of the failure mode I've seen.  On several occasions both in my own work and in supporting collaborators, I've seen the OpenCoarrays installation progress without error or warning and the resulting installation produces executable programs without warning, but the simplest of executable programs fails to launch MPI correctly.  Your suggestion to build and run a simple MPI hello world program is spot on.  In such situations, we could either terminate the build with an error message or attempt to recover by finding a working MPI installation and offering the user the option to build with it.  

[...]

PR #253, opened 2016-11-15
Better cmake mpi passing #39

zbeekman:
This (#39), along with #79 are blocking #22.

Now through the magic of `find_package(MPI)` one need not pass in the MPI compiler wrapper to CC and FC, but simply passing FC=gfortran, CC=gcc-6 should do the trick.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ env:
     - MPICH_DIR="$HOME/.local/usr/mpich"
     - MPICH_BOT_URL_HEAD="https://github.com/sourceryinstitute/opencoarrays/files/452136/"
     - MPICH_BOT_URL_TAIL="mpich-3.2_2.yosemite.bottle.1.tar.gz"
+    - FC=gfortran-6
+    - CC=gcc-6
+    - CXX=g++-6
 
 matrix:
   include:
@@ -84,9 +87,6 @@ before_install:
       [[ -d "$CACHE/bin" ]] || mkdir -p "$CACHE/bin"
       [[ -d "$MPICH_DIR" ]] || mkdir -p "$MPICH_DIR"
       export PATH="$CACHE/bin:$PATH"
-      export FC=gfortran-6
-      export CC=gcc-6
-      export CXX=g++-6
       $FC --version
       $CC --version
       $CXX --version
@@ -103,8 +103,6 @@ install:
         [[ "$(brew ls --versions $pkg)" ]] || brew install --force-bottle $pkg
         brew outdated $pkg || brew upgrade --force-bottle $pkg
       done
-      export FC=gfortran-6
-      export CC=gcc-6
       if ! [[ "$(brew ls --versions mpich)" ]] && [[ "X$BUILD_TYPE" != "XInstallScript" ]]; then
         wget ${MPICH_BOT_URL_HEAD}${MPICH_BOT_URL_TAIL}
         brew install --force-bottle ${MPICH_BOT_URL_TAIL}
@@ -114,8 +112,6 @@ install:
         mpif90 --version
         mpicc --version
         cmake --version
-        export FC=mpif90
-        export CC=mpicc
       elif [[ "X$BUILD_TYPE" = "XInstallScript" ]]; then # uninstall some stuff if present
         [[ "$(brew ls --versions cmake)" ]] && brew rm cmake || true
         [[ "$(brew ls --versions mpich)" ]] && brew rm mpich || true
@@ -143,8 +139,6 @@ install:
       mpif90 --version
       mpicc --version
       cmake --version
-      export FC=mpif90
-      export CC=mpicc
     fi
     set +o errexit
 
@@ -152,8 +146,6 @@ script:
   - |
     set -o errexit
     if [[ "X$BUILD_TYPE" = "XInstallScript" ]]; then
-      export FC=gfortran-6
-      export CC=gcc-6
       [[ -d "$HOME/opt" ]] || mkdir "$HOME/opt"
       [[ -d "$HOME/bin" ]] || mkdir "$HOME/bin"
       ln -fs "$(which gfortran-6)" "$HOME/bin/gfortran"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,21 @@ if(gfortran_compiler)
   unset(CMAKE_REQUIRED_FLAGS)
 endif()
 
+
+#----------------------------------------------------------------------------
+# Find MPI and set some flags so that FC and CC can point to gfortran and gcc
+#----------------------------------------------------------------------------
+find_package(MPI REQUIRED)
+
+set(CMAKE_C_COMPILE_FLAGS ${CMAKE_C_COMPILE_FLAGS} ${MPI_C_COMPILE_FLAGS})
+set(CMAKE_C_LINK_FLAGS ${CMAKE_C_LINK_FLAGS} ${MPI_C_LINK_FLAGS})
+set(CMAKE_Fortran_COMPILE_FLAGS ${CMAKE_Fortran_COMPILE_FLAGS} ${MPI_Fortran_COMPILE_FLAGS})
+set(CMAKE_Fortran_LINK_FLAGS ${CMAKE_Fortran_LINK_FLAGS} ${MPI_Fortran_LINK_FLAGS})
+include_directories(BEFORE ${MPI_C_INCLUDE_PATH} ${MPI_Fortran_INCLUDE_PATH})
+
+#-------------------------------
+# Recurse into the src directory
+#-------------------------------
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src)
 
 add_subdirectory(src)
--- a/src/mpi/CMakeLists.txt
+++ b/src/mpi/CMakeLists.txt
@@ -1,6 +1,14 @@
 set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/mod)
 
-find_package(MPI REQUIRED)
+if (NOT MPI_C_FOUND)
+  find_package(MPI REQUIRED)
+
+  set(CMAKE_C_COMPILE_FLAGS ${CMAKE_C_COMPILE_FLAGS} ${MPI_C_COMPILE_FLAGS})
+  set(CMAKE_C_LINK_FLAGS ${CMAKE_C_LINK_FLAGS} ${MPI_C_LINK_FLAGS})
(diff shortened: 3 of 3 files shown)


zbeekman:
:zap:

rouson:
@zbeekman

i'm wondering if the recent CMake changes have broken the installation script.  I just attempted to run it on a new platform using
```
./install.sh -p mpich
./install.sh -M prerequisites/mpich/3.1.4
```
The first command above correctly installed mpich in the location specified in the second command.  The second command identifies that this platform (a KNL box at Oregon) has a CMake version that is too old so it attempts to install CMake, which progresses really far (ran for roughly 20 minutes) and then fails with the error message below indicating that CMAKE_C_Compiler and CMAKE_Fortran_Compiler have not been set. I've never seen this before, which is why I'm suspecting it's related to this merge.   I'll try reverting to a commit before the merge, but it's going to be  a bit of a pain given how long the build goes before failing.
``` 
------ snip --------
-- Installing: /home/users/rouson/opencoarrays/prerequisites/installations/cmake/3.4.0/share/cmake-3.4/completions/cmake
-- Installing: /home/users/rouson/opencoarrays/prerequisites/installations/cmake/3.4.0/share/cmake-3.4/completions/cpack
-- Installing: /home/users/rouson/opencoarrays/prerequisites/installations/cmake/3.4.0/share/cmake-3.4/completions/ctest
2016-11-23 18:02:56 UTC [     info] popd
~/opencoarrays/prerequisites
2016-11-23 18:02:56 UTC [     info] Cleaning up. Done
install.sh: Installation successful.

install.sh: cmake is in /home/users/rouson/opencoarrays/prerequisites/installations/cmake/3.4.0/bin 

install.sh: export CMAKE=/home/users/rouson/opencoarrays/prerequisites/installations/cmake/3.4.0/bin/cmake
~/opencoarrays/prerequisites/builds/opencoarrays/1.7.5 ~/opencoarrays/prerequisites
2016-11-23 18:02:58 UTC [     info] Configuring OpenCoarrays in /home/users/rouson/opencoarrays/prerequisites/builds/opencoarrays/1.7.5 with the command:
2016-11-23 18:02:58 UTC [     info] CC="prerequisites/installations/mpich/3.1.4//bin/mpicc" FC="prerequisites/installations/mpich/3.1.4//bin/mpif90" /home/users/rouson/opencoarrays/prerequisites/installations/cmake/3.4.0/bin/cmake "/home/users/rouson/opencoarrays" -DCMAKE_INSTALL_PREFIX="/home/users/rouson/opencoarrays/prerequisites/installations/"
CMake Error at prerequisites/installations/cmake/3.4.0/share/cmake-3.4/Modules/CMakeDetermineCCompiler.cmake:57 (message):
  Could not find compiler set in environment variable CC:

  prerequisites/installations/mpich/3.1.4//bin/mpicc.
Call Stack (most recent call first):
  CMakeLists.txt:10 (project)


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_Fortran_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
See also "/home/users/rouson/opencoarrays/prerequisites/builds/opencoarrays/1.7.5/CMakeFiles/CMakeOutput.log".
2016-11-23 18:02:58 UTC [     info] Cleaning up. Done
```

rouson:
Oops... I think the error is because I gave a relative path rather than absolute path in the second command.  I'll restart the build and then report the results.

zbeekman:
Which machine, Grover? Disk IO is super slow on Grover, so you can make a temp directory in `/dev/shm` and work from there, it should be much faster. You probably want to set the installation destination back to your home directory, however, because you have to remember to delete everything from `/dev/shm ` when your done. 

I can take a look if you make it g+rwX I can take a look too.

rouson:
yes, it's grover.   hmm... i already launched the new build and I'm pretty sure my mistake is what I wrote above so I think I'll just let it continue to progress in the background while I do other things.  however, in the future, I'll do as you suggest and work in `/dev/shm` but install to my home directory.  

I just executed `chmod u+rwX opencoarrays`.  FYI, before starting, my `.bashrc` file executes something like 
```
if [[ ${HOSTNAME} == "grover" ]]; then
  module load gcc/6.1
fi
```

zbeekman:
OK, there's also CMake 3.3.1 available as a module, so you may be able to save some time by loading that up so you're not rebuilding CMake...

rouson:
My recollection is that CMake 3.4.0 was the first version that could correctly identify the version numbers of Fortran compilers so I think that's why it's set as the minimum. 

I was correct in identifying my previous mistake.  With that corrected, CMake 3.4.0 builds with no problem using
```
./install.sh -M prerequisites/mpich/3.1.4
```
It appears to me that OpenCoarrays installs correctly and the few tests I've tried running manually pass, but for some reason ctest is unable to find the tests.  Maybe I'm exhibiting a knee-jerk tendency, but I'm again wonder if this merge is the issue.  I'm using
```
cd prerequisites/build/opencoarrays/
ctest
```
The bottom of the resulting output is below:
```
---- snip -----
27/28 Test #27: asynchronous_hello_world .........***Not Run   0.00 sec
      Start 28: coarray_navier_stokes
Could not find executable MPIEXEC-NOTFOUND
Looked in the following places:
MPIEXEC-NOTFOUND
MPIEXEC-NOTFOUND
Release/MPIEXEC-NOTFOUND
Release/MPIEXEC-NOTFOUND
Debug/MPIEXEC-NOTFOUND
Debug/MPIEXEC-NOTFOUND
MinSizeRel/MPIEXEC-NOTFOUND
MinSizeRel/MPIEXEC-NOTFOUND
RelWithDebInfo/MPIEXEC-NOTFOUND
RelWithDebInfo/MPIEXEC-NOTFOUND
Deployment/MPIEXEC-NOTFOUND
Deployment/MPIEXEC-NOTFOUND
Development/MPIEXEC-NOTFOUND
Development/MPIEXEC-NOTFOUND
Unable to find executable: MPIEXEC-NOTFOUND
28/28 Test #28: coarray_navier_stokes ............***Not Run   0.00 sec

0% tests passed, 28 tests failed out of 28

Total Test time (real) =   0.21 sec

The following tests FAILED:
	  1 - initialize_mpi (Not Run)
	  2 - register (Not Run)
	  3 - register_rename_me (Not Run)
	  4 - register_rename_me_too (Not Run)
	  5 - allocate_as_barrier (Not Run)
	  6 - allocate_as_barrier_proc (Not Run)
	  7 - get_array (Not Run)
	  8 - get_self (Not Run)
	  9 - send_array (Not Run)
	 10 - get_with_offset_1d (Not Run)
	 11 - whole_get_array (Not Run)
	 12 - strided_get (Not Run)
	 13 - co_sum (Not Run)
	 14 - co_broadcast (Not Run)
	 15 - co_min (Not Run)
	 16 - co_max (Not Run)
	 17 - syncall (Not Run)
	 18 - syncimages (Not Run)
	 19 - syncimages2 (Not Run)
	 20 - duplicate_syncimages (Not Run)
	 21 - co_reduce (Not Run)
	 22 - syncimages_status (Not Run)
	 23 - simpleatomics (Not Run)
	 24 - hello_multiverse (Not Run)
	 25 - coarray_burgers_pde (Not Run)
	 26 - co_heat (Not Run)
	 27 - asynchronous_hello_world (Not Run)
	 28 - coarray_navier_stokes (Not Run)
Errors while running CTest
```

zbeekman:
cmake should have errored because it's not finding MPI. This also means that it should not have built open coarrays. I'm very puzzled by this behavior and trying to reproduce right now. Not finding MPI should have caused the same issues in previous versions too. I'm debugging on grover right now...

rouson:
Thanks for looking into this.  I'm about to be offline for a couple of hours, but I'll be glad to assist by phone if I can help and I'm up for a screens-share around 3 PM my time if that's not too late for you.

zbeekman:
Hi Damian, I've found the source of this issue: This behavior happens when `FC=/full/path/to/mpif90` and `CC=/full/path/to/mpicc` is used but the MPI implementation is not in a normal location or on the users path. I think this behavior has been present for a while now, and we've never run into it before because:

1) The install.sh script tests on Travis-CI do not invoke CTest at all, which I plan to fix ASAP (the cmake based tests *do* actually use CTest...)
2) In my testing CMake was always able to find the MPI implementation because I had it installed in a "normal" spot

I think the best way to correct this behavior is to use the `HINTS` and `PATHS` option to `find_package(MPI ... )`. We can try to compute the full path to `mpicc`/`mpif90` dynamically if we detect that it is passed, and we can add `prerequisites/installations/mpich/x.y.z/` to the `PATHS` option. This should cause CMake to pick up MPI whenever `install.sh` is used.

rouson:
@zbeekman 

I agree that you're on the right path and I'm not sure what the best solution is, but I'm a little worried about the fact that we can't anticipate all the possible locations.  For example, a style that I picked up from @sfilippone (and which seems especially important with OpenCoarrays) is to add the compiler and compiler version to the installation path, in which case the default installation path for MPICH 3.1.4 build by GCC 6.1.0 would be something like
```
${OPENCOARRAYS_SRC_DIR}/prerequisites/installations/mpich/3.1.4/gnu/6.1.0
```
and although I use "gnu" in the path name (another style I picked up from @sfilippone) because it matches the corresponding module name, I see that on grover, "gcc" is the module name instead of "gnu". 

As an aside that might or might not help to know, the paths that `install.sh` uses for prerequisites get written to a `setup.sh` script in the OpenCoarrays installation directory.  I source that script in my .bashrc file.

zbeekman:
@rouson 

We don't need to anticipate all the locations MPICH could be. Here's why:

- CMake is pretty smart on its own. If the MPI wrappers and `mpirun` are on the users `PATH` (probably a safe assumption for most users) there are no issues, no matter where they install MPI
- Testing for MPI wrappers being passed as "compilers" via `CC` and `FC` will allow us to tell CMake to go look in the directory above wherever those are coming from via the `HINTS` directive (I was falsely under the impression that CMake already does something like this, not hard to implement)
- Failing that, the `PATH` directive can point to the default MPICH install location in from `./install.sh`.

The installation procedure you outlined above doesn't appear to put `mpich/x.y.z/bin` directory on the users path, so `find_package(MPI)` was not able to succeed, BUT the MPI wrapper scripts were passed explicitly, so that didn't cause it to fail either. In fact, in the past, it's possible that the MPI being used to build OpenCoarrays, and the version used at run time may have been able to be different because of this. Perhaps this is a possible source of some/all of the behavior discussed in #246... between these fixes, and some more testing/introspection at CMake/configure time, I think there's a reasonable chance #246 gets fixed with these updates

zbeekman:
Note to self: the FindMPI module (and find modules in general) behave differently than find_package, therefore the variable that needs tweaking is `MPI_HOME` not `PATHS` or `HINTS`.

Issue #254, opened 2016-11-22
sendget with strides is failing


zbeekman:
reported by @vehre here: https://github.com/sourceryinstitute/opencoarrays/pull/250#issuecomment-261204819

zbeekman:
https://github.com/sourceryinstitute/opencoarrays/files/597019/stridedsendget.f90.gz

PR #255, opened 2016-11-23
Bugfix strided sendget

afanfa:
Hi,
this patch fixes issue #254 and adds the testcase written by @vehre .


rouson:
@afanfa, please ensure that the test case is well documented.   The last version I saw contained no comments other than several commented `error stop` statements.  In addition to a comment explaining the purpose of the test, let's either remove the `error stop` statements or add a comment that explains why they are commented out or (preferably) block them with conditional compilation using a descriptive variable name that indicates when they should be used.  I'll be glad to help with this process if you'd like to do a screen share.  If so, I'll be available today after about 3 PM.

vehre:
@rouson Why needs the "testcase" documentation? Documenting the library's source is far more important. The testcase will be rather seldom maintained, but the library's source is under constant maintenance. I agree, that a description should got at the start of the testcase's file, but the code is rather self explanatory.

src/mpi/mpi_caf.c
```diff
@@ -912,37 +912,42 @@ PREFIX (sendget) (caf_token_t token_s, size_t offset_s, int image_index_s,
...
                                       - dest->dim[j].lower_bound + 1))
                 * dest->dim[j]._stride;
               extent = (dest->dim[j]._ubound - dest->dim[j].lower_bound + 1);
               stride = dest->dim[j]._stride;
+	      tot_ext *= extent;
             }
 
+	  array_offset_dst += (i / tot_ext) * dest->dim[rank-1]._stride;
+	  
 	  extent = (dest->dim[rank-1]._ubound - dest->dim[rank-1].lower_bound + 1);
```
vehre: This line IMO can be removed.

src/mpi/mpi_caf.c
```diff
@@ -912,37 +912,42 @@ PREFIX (sendget) (caf_token_t token_s, size_t offset_s, int image_index_s,
...
                                          - src->dim[j].lower_bound + 1))
                     * src->dim[j]._stride;
                   extent = (src->dim[j]._ubound - src->dim[j].lower_bound + 1);
                   stride = src->dim[j]._stride;
+		  tot_ext *= extent;
                 }
 
-	      extent = (src->dim[rank-1]._ubound - src->dim[rank-1].lower_bound + 1);
-              array_offset_sr += (i / extent) * src->dim[rank-1]._stride;
+	      /* extent = (src->dim[rank-1]._ubound - src->dim[rank-1].lower_bound + 1); */
```
vehre: Can be removed, too.

afanfa:
All the changes look good to me.

rouson:
> On Nov 24, 2016, at 3:00 AM, Dr. Andre Vehreschild <notifications@github.com> wrote:
> 
> @rouson <https://github.com/rouson> Why needs the "testcase" documentation?
> 
I don’t understand the resistance to documentation.  It takes no more time to write a few comments than it took to write this response.
> Documenting the library's source is far more important.
> 
That doesn’t make documenting tests unimportant.  It’s a crime against humanity (namely against other developers) to write a file that doesn’t have even a single comment to explain the code’s purpose.
> The testcase will be rather seldom maintained, but the library's source is under constant maintenance.
> 
Every time the test fails, someone has to read the test to see what failure it’s identifying.  For anyone other than the author, that likely means reading the whole file when in fact a comment or two could explain the purpose.

> I agree, that a description should got at the start of the testcase's file, but the code is rather self explanatory.
> 

Code is always self-explanatory to the author.  Even the simplest of code is not necessarily self-explanatory to another reader — especially when magic numbers appear as is almost always the case in parallel code. Image 1 does x… Image 3 does y… Is there anything special about choosing those image numbers or would any two images suffice?  A simple comment that says “test what happens when one image does x while another image does y” will hopefully make it more clear that the specific image numbers do not matter.

I used to believe in the philosophy “write code that comments itself.”  After several years of teaching programming to total of over 500 people in about 25 short courses and 4 graduate courses at two universities plus having gotten feedback about my book, which adopted the “comments itself” philosophy, I eventually concluded that the philosophy is a lie.  Code never comments itself.  That’s just something that we tell ourselves to free ourselves from the burden of writing comments.

Damian

CMakeLists.txt
```diff
@@ -213,6 +213,7 @@ if(opencoarrays_aware_compiler)
   add_mpi_test(get_with_offset_1d 2 ${tests_root}/unit/send-get/get_with_offset_1d)
   add_mpi_test(whole_get_array 2 ${tests_root}/unit/send-get/whole_get_array)
   add_mpi_test(strided_get 2 ${tests_root}/unit/send-get/strided_get)
+  add_mpi_test(strided_sendget 4 ${tests_root}/unit/send-get/strided_sendget)
```
zbeekman: Is there any harm in over subscribing?

I've found that often over subscribing can find race conditions and other parallel programming errors that might not otherwise show up.

For OpenMPI, the tests will run on the number of available cores, because current OpenMPI implementations are unreliable (at least on our virtualized testing infrastructure)

src/mpi/mpi_caf.c
zbeekman: I'll defer to you and @afanfa here

src/tests/unit/send-get/strided_sendget.f90
```diff
@@ -0,0 +1,67 @@
...
+  end if
+
+  sync all
+  if (this_image() == 1) then
+    dstvec(:)[3] = srcvec(:)[2]
+    dstmat(3,:)[3] = srcvec(:)[2]
+  end if
+
+  sync all
+  if (this_image() == 3) then
```
zbeekman: Is there a way to abstract out image numbers 1, 2 and 3? i.e. is there a way to ensure that the test would at least pass (if not actually test for correctness) on 1 or 2 images?

CMakeLists.txt
rouson: I agree: oversubscribing is a great way to identify problems, including the [bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78505) that I just submitted on gfortran last night.

vehre:
I have added a pull request (#257) to add the documentation to the testcase.

> On Nov 24, 2016, at 3:00 AM, Dr. Andre Vehreschild ***@***.***> wrote:
>
 > > @rouson <https://github.com/rouson> Why needs the "testcase" documentation?
> I don’t understand the resistance to documentation.  It takes no more time to write a few comments than it took to write this response.

Ok, I should have phrased this "Why needs the example I submitted to show the issue (never thought as a testcase) more documentation than the other testcases in this directory?" By adding the requested comments the line count of comments in the directory rises by one third from (57 to 95 lines of comment; not counting the old_get_array_test.f90). What I criticize is the difference of criteria to accept a testcase for me and others submitting tests.

> That doesn’t make documenting tests unimportant.  It’s a crime against humanity (namely against other developers) to write a file that doesn’t have even a single comment to explain the code’s purpose.

My code was made a testcase, but not by me. Why I have to spend documenting a sample showing an issue, is lost to me.

> > The testcase will be rather seldom maintained, but the library's source is under constant maintenance.
> Every time the test fails, someone has to read the test to see what failure it’s identifying.  For anyone other than the author, that likely means reading the whole file when in fact a comment or two could explain the purpose.

Well, whether its reading the code or the comments might be the same effort. Given that most of the time the code is more current than the comments and the comments not seldomly are wrong. Therefore I usually do not comment thoroughly, but only the outstanding necessities.

> zbeekman 18 hours ago Sourcery Institute member
>
> Is there a way to abstract out image numbers 1, 2 and 3?

The test in pull request #257 should address this issue.

> i.e. is there a way to ensure that the test would at least pass (if not actually test for correctness) on 1 or 2 images?

Well, the issue can be, that sendget uses optimization when its source- or dest-image is this_image. To test the full sendget functionality, three images are required, one providing the source, one the destination memory and one acting as master doing the communication.

CMakeLists.txt
vehre: Oversubscribing in general is not doing harm, but it does not test whether the test runs with the smallest number of images it is expected to run with. So, to do this correctly one would need to set up tests:
- with less than the **n**umber **o**f **i**mages **r**equired by the test (noir),
- exactly the noir,
- with noir + 1 image, and
- with significantly more images than noir.

So, do you really want to do this for every test?

src/mpi/mpi_caf.c
vehre: Oh, sorry, I haven't used githubs review caps yet. So my comment went to the wrong line. Not the line starting with `array_offset_dst +=` should be removed, but the line starting with `extent = `. Extent is not read after set here anymore.

src/mpi/mpi_caf.c
vehre: Same here, the commented line can be remove.

src/mpi/mpi_caf.c
zbeekman: @afanfa can you please confirm this?

@vehre can you please elaborate on why it is OK to remove this? I'm not  familiar enough with the library internals to immediate see why this is safe to be removed.

vehre: Because 'extent' is set to 1 thereafter and no read occurs inbetween.

CMakeLists.txt
zbeekman: Looking at the details of this particular test, I agree that we should limit to 3 images... I'll change this

src/mpi/mpi_caf.c
zbeekman: I see it now, thanks

zbeekman:
I think this has addressed everyones comments and concerns. @afanfa @vehre @rouson please confirm that everything looks good.

vehre:
I feel that all issues have been addressed.

zbeekman:
LGTM ... I'll wait a few more hours for the others to weigh in.

zbeekman:
LGTM, once the CI checks pass... Travis CI is still having OS X backlog issues

PR #257, opened 2016-11-25
Fix testcase for strided_sendget.

vehre:
Thoroughly document unit/send-get/strided_sendget.f90.
Use only the needed number of images for the testing strided_sendget.


zbeekman:
LGTM

I think I may find a way to make this pass---or at least not fail---when we can't run with 3 or more images

vehre:
Well, when we can't run with 3 or more images, then the whole testsuite has an issue. Why would you want to spend time on this? Opencoarrays is about parallel execution, when there is no opportunity to exploit parallelism there no need to run any test. I don't understand why this bothers.

rouson:
@vehre I believe Zaak is simply referring to situations in which someone is running the test in a limited environment for which running more images doesn't make much sense.  One common scenario would be running on a dual-core laptop.  We currently have an open issue that only occurs when the cores are oversubscribed.  Therefore, anyone running on a dual-core laptop cannot simultaneously run with 3 images and expect to pass all tests.

zbeekman:
Yes, I think I have added some logic, for the time being to the test itself
along the lines of `if (num_images() < 3) then ; write(*,*) 'Skipping test
due to insufficient cores' ; write(*,*) 'Test passed.' ; stop 0 ; end if`
or something to that effect. The better approach would be to adjust the
CMake macro so that we have a requested images and a minimum images
argument so that tests requiring resources that are not available will just
be skipped. That would be the correct way to do it, but that's going on the
someday task list for now.

vehre:
Is it because of the not synchronizing ALLOCATE(, SOURCE= )? The fix for that is in review.

zbeekman:
@vehre in review with GCC/GFortran? Do you have a link to the thread/PR?

vehre:
78505

zbeekman:
Awesome, thanks for taking care of that! [GCC PR 78505](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78505) patch here: https://gcc.gnu.org/ml/fortran/2016-12/msg00012.html

PR #258, opened 2016-11-30
More robust MPI handling/finding

zbeekman:
Untested, but this *should* behave better than the previous
 implementation and should try really hard to "do what you mean" ™️ 

I'll also try to address #246 as best we can... just do it through checking the .mod file compatibility... readelf is not something I want to deal with, and strings -a doesn't work because everyone uses strip before shipping binaries.

@jerryd mentioned that you might be able to unzip .mod files and get the GFortran version, but I haven't been able to do this, despite a few tries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ else()
 endif()
 
 if(gfortran_compiler)
+  set(OLD_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
   set(CMAKE_REQUIRED_FLAGS "-fcoarray=single -ffree-form")
 endif()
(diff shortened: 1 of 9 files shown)


zbeekman:
I cancelled a bunch of CI tests for individual commits since Travis-CI has a huge OS X backlog

I have also tested this on Debian 24 in every conceivable way, and everything looks kosher. 

@jerryd:  Please try test driving this branch on your Debian machine to ensure that I haven't missed something. I'll wait the usually 24 hours before self-approving unless I hear back sooner.

zbeekman:
LGTM

zbeekman:
@jerryd confirmed over Gchat that this fixed the issues he's experienced on Fedora.

PR #262, opened 2016-12-01
Update caffeinate-opencoarrays from master

zbeekman:
These branches have gotten out of sync and have some merge conflicts.

Let's create this PR to inspect and resolve them

PR #272, opened 2016-12-05
Update Caffeinate-opencoarrays from master--up to allocatable component work supersedes #264

zbeekman:
OK, supersedes and closes #262 

As an aside, I finally figured out the correct way to update one protected branch from another...


zbeekman:
@rouson not urgent, but I probably won't merge this until you take a look. Essentially the point is to get some changes (before that allocatable components work started) into caffeinate-opencoarrays from master. This will make our lives easier when we then have to merge in the allocatable components work and the changes we're making to the `caf` and `cafrun` scripts.

rouson:
LGTM

Issue #269, opened 2016-12-04
Update install.sh to use FC=gfortran, CC=gcc, MPI_HOME=/path/to/mpi/if/specified/or/installed


zbeekman:
I think it might be wise to update the install.sh scripts to pass FC=gfortran, CC=gcc rather than mpif90 and mpicc etc. The most complicated part about this is that I need to do some testing and investigation of how to handle the logic surrounding MPI. I think that if `install.sh` either:

 1. Installs mpich as part of the build, or
 2. Receives a user flag setting the MPI location

Then it should automagically "just do the right thing" :tm: *IF* we also set MPI_HOME environment variable. If the script doesn't do the install of MPICH, and the user (or script elsewhere) doesn't explicitly specify an MPI to use then CMakes FindMPI should pick it up from either:

 1. The users path via it's searching for compiler wrappers and `mpirun` etc.
 2. environment modules (gnu modules) setting the MPI_HOME variable explicitly
 3. other crazy and complicated magic from FindMPI. When CMake fails on this step it also outputs info so that the user can specify via the command line or `ccmake` or `cmake-gui` interfaces the explicit location of where to find MPI.

I think it *should* be fairly straightforward to implement this, but I need to dig into the install script to assert this with any confidence first....

zbeekman:
@rouson Currently I'm having trouble following variable scoping... I have a WIP [patch](https://github.com/sourceryinstitute/opencoarrays/compare/topic-%23269-FC-gfortran) that I think will work for build_opencoarrays.sh, but I can't figure out the correct way to get FC and CC into build_opencoarrays.sh. Lets talk about this, or perhaps you can do it for me or tell me how to do it?

Here is my idea (WIP, fails to build/test): https://github.com/sourceryinstitute/opencoarrays/compare/topic-%23269-FC-gfortran but currently complains due to lack of FC and CC.... I.E. they must be explicitly set if the user passes them on the command line via -f and -c, or they must point to the gcc that was installed as part of the build.

n.b. CC could in theory point to a different C compiler

PR #274, opened 2016-12-12
Topic #269 fc gfortran

rouson:
Replaces `find_or_install mpich` invocation with `find_or_install gcc."  Also, this pull request improves the installation script stack management and testing.


zbeekman:
@rouson looking like you're making some great progress, and adding some unit tests for the shell scripts, which is great!

It looks like `./install.sh` [exits early](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/183189174#L229) on OS X and linux when invoked as

```
./install.sh --yes-to-all -i "$HOME/opt/opencoarrays" -j 4 -f "$HOME/bin/gfortran" -c "$HOME/bin/gcc" -C "$HOME/bin/g++"
```

Also, it seems that CMake/CTest is [appending an extra `.sh`](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/183189176#L685) for the shell script tests. This looks quite easy to test.

The only reason this branch is out of sync is because of some documentation fixes I've made, and those shouldn't conflict with anything we're doing here so a simple git pull origin master should do the trick to sync them up.

AddInstallationScriptTest.cmake
```diff
@@ -0,0 +1,25 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/${path}/${name}-usage
+    ${CMAKE_CURRENT_BINARY_DIR}/${path}/${name}-usage
+    COPYONLY
+  )
+  set(harness "${CMAKE_BINARY_DIR}/staging/test-${name}.sh")
+  install(
+      FILES "${harness}"
+      PERMISSIONS WORLD_EXECUTE WORLD_READ WORLD_WRITE OWNER_EXECUTE OWNER_READ OWNER_WRITE GROUP_EXECUTE GROUP_READ GROUP_WRITE
+      DESTINATION ${CMAKE_BINARY_DIR}/${path}
+  )
```
zbeekman: Do we want to *install* the shell script tests? Usually this sort of thing is just run at build time, along with unit tests and not installed....

AddInstallationScriptTest.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+  # Copy the source to the binary tree 
+  configure_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/${path}/${name}
+    ${CMAKE_CURRENT_BINARY_DIR}/${path}/${name}
+    COPYONLY
+  )
+  configure_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/${path}/${name}-usage
+    ${CMAKE_CURRENT_BINARY_DIR}/${path}/${name}-usage
```
zbeekman: If you pass in `name` with suffix `.sh` then usage is going to be `${name}.sh-usage` ...

AddInstallationScriptTest.cmake
```diff
@@ -0,0 +1,25 @@
+macro(add_installation_script_test name path)
```
zbeekman: What if we run `get_filename_component(local_name ${name} NAME_WE)` so as to always strip off any training `.sh` if it is passed in to the macro, and then explicitly add it back in the macro? This has a pro and a con

**pro**: You can pass in the script name with or without the file extension/suffix

**con**: If you have scripts that don't end in `.sh` then when you write out the files below you will be removing whatever suffix was there, if any, and adding `.sh`.

You could get rid of the con by also calling `get_filename_componenet(orig_suffix ${name} EXT)` and then adding that back: `${local_name}${orig_suffix}`

Perhaps it might make sense to also sanitize the path with `get_filename_component(path_var ${path} DIRECTORY)` but I think we'd really have to be asleep at the wheel to pass in a file there... so that last suggestion is probably overkill

zbeekman: @rouson if you're willing, I can push commits to this branch to implement this.

AddInstallationScriptTest.cmake
zbeekman: This is actually the cause, AFAICT of the failures I'm seeing on Travis-CI. This line enforces the requirement of CTest *AFTER* `make install` which is highly unusual, since the canonical order is something like:

```
cmake|configure
make
make check|make test|ctest
make install
```

I have what I believe to be a good work around.

rouson:
@zbeekman 
Feel free to fix the issues you've mentioned above however you see fit.  I'll focus on editing `find_or_install.sh` to launch `CMake with CC=gcc FC==gfortran cmake ...` as we discussed.  I assume you won't need to modify `find_or_install.sh` so there is little chance of a conflict between our edits.

zbeekman:
> Feel free to fix the issues you've mentioned above however you see fit. I'll focus on editing find_or_install.sh to launch CMake with CC=gcc FC==gfortran cmake ... as we discussed. I assume you won't need to modify find_or_install.sh so there is little chance of a conflict between our edits.

sounds good. Push up any changes you may once you're happy with them. I'll rebase any changes on top of yours so that git pulls should be easy fast forward merges.

zbeekman:
Also, FYI, as soon as we're ready to merge this, I'll probably interactively rebase it to clean up the commit history with some squashed commits, etc. so that there aren't any work in progress (WIP) commits. (Yes, I was the one who introduced the non-functional commit.)

zbeekman:
@rouson I've added my changes to CMake to add the shell script tests... Right now all of the CI tests pass, but I think you said that there was more work to do, no? IIRC there are some other cases that might not be correctly handled in terms of getting FC and CC into [`build_opencoarrays.sh`](https://github.com/sourceryinstitute/opencoarrays/pull/274/files#diff-1b6698647bafaa08c5f23797022f4929)

prerequisites/install-functions/build_opencoarrays.sh
```diff
@@ -10,18 +10,27 @@ build_opencoarrays()
...
+  info "Configuring OpenCoarrays in ${PWD} with the command:"
+  info "CC=\"${CC}\" FC=\"${FC}\" $CMAKE \"${opencoarrays_src_dir}\" -DCMAKE_INSTALL_PREFIX=\"${install_path}\" -DMPI_C_COMPILER=\"${MPICC}\" -DMPI_Fortran_COMPILER=\"${MPIFC}\""
+  CC="${CC}" FC="${FC}" $CMAKE "${opencoarrays_src_dir}" -DCMAKE_INSTALL_PREFIX="${install_path}" -DMPI_C_COMPILER="${MPICC}" -DMPI_Fortran_COMPILER="${MPIFC}"
+  info "Building OpenCoarrays in ${PWD} with the command make -j${num_threads}"
+  make "-j${num_threads}"
+  if [[ ! -z ${SUDO:-} ]]; then
+    printf "\nThe chosen installation path requires sudo privileges. Please enter password if prompted.\n"
+  fi
+  info "Installing OpenCoarrays in ${install_path} with the command ${SUDO:-} make install"
+  ${SUDO:-} make install
```
zbeekman: This looks perfect! Let me know if/when you think it's complete and ready to merge

rouson: LGTM 

Thanks for the quick feedback.  I think it's ready.  I tested the following scenarios:
- In the VM with -f/-c/-C arguments pointing to GCC 7 dated 20161112.  
- In MacOS with GCC 6.2.0 
    - with -f/-c/-C arguments.
    - with no arguments
- In MacOS with `FC=pgortran CC=pgcc ./install.sh` just to see if it went for the head fake ⛹️, which it did not. 😉 

All 29 tests pass in MacOS.  In the VM, 29 of 30 tests pass with the one previously observed failure being the get_self test.  I'm going o launch a new build of the current trunk in the VM tonight to see if the one test failure remains, but that's unrelated to this pull request so merge away!  (assuming the Travis-CI tests pass)

I also tried

zbeekman: Great! I think you need to issue the magic approval comment from a comment in the "conversation" tab, and not for inline discussions... no matter.

zbeekman:
LGTM via @rouson above

zbeekman:
doh! I meant to reword/squash the commit with `WIP` in it... gah. Now my `push` git hook is going to yell at me all the time... I'm strongly contemplating reverting master.... but it's probably not worth it.

PR #261, opened 2016-12-01
[Updated #259] Add interface changes needed for gcc-trunk and testcases.

vehre:
Same description as in #259.
Sorry for the noise.

Issue #260, opened 2016-11-30
Implement asynchronous allocation of components in derived typed coarrays and caf_is_present


vehre:
Pull request #259 adds the changes needed to opencoarrays/caf_mpi to "work" with current gcc-trunk again. By "work" I mean compile and run the existing tests.

My idea for testing async alloc of components:

[check_remote_alloced_comp_1.f90.txt](https://github.com/sourceryinstitute/opencoarrays/files/621958/check_remote_alloced_comp_1.f90.txt)

fails to run for two reasons:
1. in caf_register() the MPI_allocate() syncs and
2.  caf_is_present() is not implemented.

The test case can be used as a guide for implementing the features. At the moment it has some additional prints for debugging, which should be removed in the final version.

PR #259, opened 2016-11-30
Add interfaces changes needed for gcc-trunk and testcases.

vehre:
With today's commit to gcc-trunk the interface for asynchronously allocatable components in derived typed coarrays has been made availabe in gfortran. This pull request adds the necessary changes to allow compilation and running tests ok for opencoarrays again.

PR #276, opened 2016-12-14
Vehre/async alloc comp support

zbeekman:
Supersedes PR #261, #259 

@afanfa signed off on #261 but I wanted to clean it up and add some conditionals around the tests requiring GCC 7 so that they don't get built. 

`src/tests/unit/init_register/check_remote_alloced_comp_1.f90` is anticipated to fail when building with GCC 7 until #260 is finished


zbeekman:
So long as @vehre and @afanfa don't object, I'll merge this once the tests pass, since it all

looks good to me

afanfa:
No objections here. Go for it.

zbeekman:
still

looks good to me

PR #277, opened 2016-12-15
Edit caf/cafrun for FC=gfortran instead of mpif90

rouson:
Before this commit, the `caf` was invoking gfortran directly because of the new practice of passing compilers instead of MPI wrappers in`FC=gfortran cmake ..` and `cafrun` was hardwired to invoke `mpif90`. With this commit, `caf` invokes `${MPI_Fortran_COMPILER}` and `cafrun` invokes `${MPIEXEC}`.

### DIFF
--- a/src/extensions/caf-foot
+++ b/src/extensions/caf-foot
@@ -63,7 +63,7 @@ fi
 # Default to "mpif90" Fortran compiler if environment variable CAFC is empty or unset:
 CAFC=${CAFC:-mpif90}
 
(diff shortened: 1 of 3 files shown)


zbeekman:
Great! I'll take a look. I was about to start working on switching to configure_file for writing caf and cafrun, so this should be a great starting point!

zbeekman:
tested locally on OS X

👍

zbeekman:
beat me to it

Issue #279, opened 2016-12-17
check_remote_alloced_comp_1.f90 hangs in VM


rouson:
@vehre 

Test #9 [check_remote_alloced_comp_1.f90](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/tests/unit/init_register/check_remote_alloced_comp_1.f90) hangs in the current version of the virtual machine, which has a gfortran build from 2 days ago.  This appears to be a new test that you created 3 days ago.  

I just uploaded the latest version of the virtual machine today.  It has GCC compilers dated 20161215.

vehre:
@zbeekman, @afanfa this test should only be active, when #260 is done.

rouson:
@vehre In that case, 

1. The feature should not be enabled on the OpenCoarrays master branch, or
2. There should be a compiler error message tha blocks use of the broken features,  or 
3. OpenCoarrays should give a generate a runtime error message whenever a code tries to access the broken feature.  

listed in order from most preferred to least preferred solution.  Up until this allocatable components work, we've held to a very high standard of not knowingly letting broken features hit the master branch.  This causes a great deal of confusion and wasted time.  Just yesterday, I uploaded a new copy of the virtual machine with a GCC trunk build from 20161215 and notified my class of it because it includes several fixes to bugs my students and I discovered during the course.   Once one of the above happens, I'll have to rebuild GCC and/or OpenCoarrays, then export and post a new virtual machine, and notify everyone.  

I tested OpenCoarrays before posting the virtual machine, but I think my local repository most have been a day or two behind master so I didn't catch this test failure until after releasing.

zbeekman:
@rouson this is entirely my fault. @vehre told me about this, and I neglected to disable it. I'll make the fix right now.

zbeekman:
to elaborate, I was short sighted in that I thought that no one would be using GCC trunk "in production" and that the `if GCC >= 7` guards around the test would be sufficient.

zbeekman:
In ~20 minutes CI tests should be done for PR #280 and we can merge the change to fix this.

@rouson 
> listed in order from most preferred to least preferred solution.

Did you mean "listed in order from least preferred to most preferred solution"?

zbeekman:
> OpenCoarrays should generate a runtime error message whenever a code tries to access the broken feature.

Some sort of quick stub routine in the OpenCoarrays library that would error stop rather than just hang would be nice.

zbeekman:
[...]

PR #281, opened 2016-12-17
set ctest to time-out on failing alloc comp test

rouson:


rouson:
@zbeekman, it does of course make sense to deactivate this test for gfortran < 7.  now that we have the ctest time-out in place, this can be active gfortran >=7.  Thanks for your help!

rouson:
I should have included "Fixes #279" in the commit message.  I'm confident this will pass so I'm going to go ahead and close that issue.

zbeekman:
You get to write commit messages for the merge too.... and I think because of waffle.io rules and the branch name that may have closed the issue too

zbeekman:
In general I like to keep them open until code hits master though

zbeekman:
One of the PR sub jobs errored for unknown reasons... I restarted it.

LGTM

rouson:
Now that all tests have passed, I'll merge this one.  Obviously this meets the requirement of review by a second developer. :)

zbeekman:
Yup and I even used the *official* way to sign off with pullapprove

vehre:
This unfortunately did not work on my system (bare metal). The test hung for ever and did not timeout even after 5 minutes waiting. I had to add:

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 20018dc..372967f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,6 +371,7 @@ if(opencoarrays_aware_compiler)
     add_mpi_test(register_alloc_comp_3 2 ${tests_root}/unit/init_register/register_alloc_comp_3)
     add_mpi_test(async_comp_alloc 6 ${tests_root}/unit/init_register/async_comp_alloc)
     # Timeout async_comp_alloc test after 3 seconds to progess past the known failure
+    set_property(TEST async_comp_alloc PROPERTY TIMEOUT 6)
     set_property(TEST async_comp_alloc PROPERTY TIMEOUT_AFTER_MATCH 3 "known failure")
   endif()
   add_mpi_test(get_array 2 ${tests_root}/unit/send-get/get_array)

to make the test timeout after at least 6 seconds.

TIMEOUT_AFTER_MATCH is not available on my system installed cmake 3.4.3. So I propose to additionally use a general timeout for the test, which is available from cmake 2.6 on. It does no harm to have a longer general timeout and will allow ctest to run on more systems.

zbeekman:
@vehre sounds good. Did CMake generate any errors/warnings when it encountered `TIMEOUT_AFTER_MATCH`? If so can you paste them here?

vehre:
Nop, cmake silently accepted the TIMEOUT_AFTER_MATCH.

zbeekman:
With CMake it is hard to follow which feature was added in which release... I guess because it's a test *property* it makes sense that if it's missing it's a noop. Let's add your code because it's a good safety net anyway.

I need to take a run at the CMake scripts with the dev warnings all turned on to find more potential issues like this.

vehre:
Yes, that's a common issue with cmake. Even the documentation lacks hints with which version a property got available.

The issue with properties is, that you can define your own. So cmake can not know whether the property is user-defined or referencing something available only in the future.

rouson:
In general, I'd rather bump the version requirements for prerequisites than design around older versions. @zbeekman, what CMake version is used in our Travis-CI builds?

The very nature of modern Fortran is such that it will be a very long time before we are able to settle on stable version requirements for our prerequisites.   I'll propose CMake 3.7.0 as the minimum sometime in the coming months (probably when we release OpenCoarrays 2.0.0, which is likely to be in the Spring when gfortran 7.1.0 releases).  CMake 3.7.0 is the first version that supports Fortran 2008 submodules. Sourcery Institute funded the submodule support in gfortran and the submodule support in CMake.  We should leverage those investments. (For example, [opencoarrays.f90](https://github.com/sourceryinstitute/opencoarrays/blob/master/src/extensions/opencoarrays.F90) would benefit greatly from refactoring to increase its modularity with submodules.) And I'd like to encourage the rest of the world to come along with us into the future.

vehre:
Please keep in mind, that older gfortran version will not get submodule support. So asking a higher cmake version for them does not pay, when it is only for the ability to use a feature of cmake that is not essential. I mean, what is the use of scanning for the regex and then restarting the timeout, or giving the whole test a slightly longer timeout? The result is the same, the test times out as long as async alloc is not implemented. When it is implemented the test will pass in any variant.

It is nice to have this message, saying that there probably is an issue, but why do you need to scan for it? When the test does not pass or times out, then it is examined. IMO this is independent of the message.

zbeekman:
> what CMake version is used in our Travis-CI builds?

@rouson [on OS X we use Homebrew's latest](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/184851115#L77), which currently appears to be 3.6.2 and but [on linux it is 3.2.3](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/184851118#L291). Perhaps a more recent version is available through the debian-sid PPA... I'll go take  a look.

zbeekman:
I agree with the sentiment that getting people to modernize is good, but I also fear alienating users if we make them update all the things... And in this particular case I agree with Andre that it is probably better to not alarm users who are running with older CMake, and ensure that the test properly fails. I can look into ways of adding extra messaging to the user that this is an expected failure.

rouson:
@zbeekman sounds good.  We can stick with CMake 3.4.0 through the 2.0.0 release.  Given that opencoarrays.F90 exists to support source-to-source transformation, we can delay the bump to CMake 3.7.0 until the OpenCoarrays 3.0.0.

zbeekman:
working on this now, but also trying to get up-to-date CMake on the Travis-CI infra

PR #284, opened 2016-12-20
Add safer timeout for GCC 7 test and compact .travis.yml

zbeekman:
See discussion at #281 for details on the CTest timeout.

Attempting to decrease verbosity of the `matrix.include` element for `.travis.yml` using YAML anchors, references and merges.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,18 +27,20 @@ matrix:
       env:
        - BUILD_TYPE="InstallScript"
          OSX_PACKAGES="gcc"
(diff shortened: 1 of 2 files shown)


zbeekman:
I originally tried to add the debian-sid PPA, but it broke everything because it tries to get gcc, glib6 etc. from debian-sid. Unless we build CMake from source, we'll have to deal with an oldish version on CI... for now at least.

zbeekman:
LGTM

PR #294, opened 2016-12-23
Event post regression when OpenCoarrays is built with GCC 7

rouson:
This pull request
* Adds a regression test for an event post failure that occurs when num_images()>2 and OpenCoarrays has been built with gfortran 7. 
* Prevents the async-hello.F90 integration test from erroneously reporting a passing test in the above scenario.
After this regression has been resolved, preprocessor conditionals in both of the above tests must be removed to facilitate reporting a passing test.


rouson:
LGTM

PR #296, opened 2016-12-25
Edit GCC download_prerequisites

rouson:
This pull request strengthens the installation scripts for building a broader range of GCC versions in a broader array of settings.  In particular, the scripts should now work with any combination of
* GCC 5, 6, and the trunk (currently 7),
* On macOS or Linux, 
* With either curl, wget, or ftp installed and in the user's PATH.

When required by the installation environment, the OpenCoarrays installation scripts modify GCC's [download_prerequisites] script to use curl or ftp instead of GCC's [hardwired wget] choice.  For example, macOS systems have `curl` installed by default but not `wget`; whereas Lubuntu Linux has `wget` installed by default but not `curl`; and old systems might only have `ftp`.  

GCC 7 changed the form of the `wget` command line inside the GCC download_prerequisites script.  The OpenCoarrays installation scripts are designed to work for versions before and after the change.  The OpenCoarrays scripts use pattern-matching to edit download_prerequisites.  Where possible, the pattern matching is sufficiently flexible to be robust in the face of future, minor changes in download_prerequisites, e.g., changes to white space or re-ordering of certain arguments.

[download_prerequisites]: https://gcc.gnu.org/viewcvs/gcc/trunk/contrib/download_prerequisites?revision=241483&view=markup
[hardwired wget]: https://gcc.gnu.org/viewcvs/gcc/trunk/contrib/download_prerequisites?revision=241483&view=markup#l173


zbeekman:
It seems that Travis-CI thought that the ctest command didn't exit: https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/186706729#L9229

This may just be cloud noise, or perhaps it could be indicative of some sort of test hanging/segfaulting like #294... I'm going to try to default to verbose ctest on Travis-CI to help us know with more certainty that the tests are actually passing...

rouson:
LGTM. All tests have passed.

Issue #293, opened 2016-12-23
Regression: event post seg faults with num_images()>2 with gfortran 7


rouson:
With gfortran 7.0.0 dated 20161215, the issue demonstrated below was causing a silent failure in the [async-hello.F90] integration test, which progressed through the printing of "Test passed." even though all images report segmentation faults caused by the `event post`.  No such problem occurs when OpenCoarrays is built with gfortran 5 or 6.  This is a very serious regression that was identified by code provided by @egiovan.  Given that gfortran 7 is expected to release in January, I hope we can fix this soon.

A pull request to add the code below in src/tests/regressions/reported in order to have a simpler demonstration of the problem than aforementioned integration test.
```
$ cat issue-293-silent-event-failure.F90 
program main
  use iso_fortran_env, only : event_type
  implicit none
  type(event_type) :: test_post[*]
  if (num_images()<3) error stop "exposing issue 293 requires num_images()>=3"
  event post(test_post)
  if (this_image()==1) print *,"Test passed."
end program

$ caf issue-293-silent-event-failure.F90 

$ cafrun -np 3 ./a.out
 Test passed.

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7fe53280c4af in ???
#1  0x4045d0 in ???
#2  0x401b91 in ???
#3  0x7fe5327f782f in ???
#4  0x401968 in ???
#5  0xffffffffffffffff in ???
#0  0x7f029cf004af in ???
#1  0x4045d0 in ???
#2  0x401b91 in ???
#3  0x7f029ceeb82f in ???
#4  0x401968 in ???
#5  0xffffffffffffffff in ???
#0  0x7f0470b044af in ???
#1  0x4045d0 in ???
#2  0x401b91 in ???
#3  0x7f0470aef82f in ???
#4  0x401968 in ???
#5  0xffffffffffffffff in ???

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 7660 RUNNING AT Sourcery-Linux-VM
=   EXIT CODE: 139
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
[...]

PR #298, opened 2016-12-29
Null array-descriptor in mpi-token causing segfault on finalize. (#293)

vehre:
This patch ensures that the array descriptor stored in the mpi-caf-token is
set to NULL consistently, when no real descriptor is stored in it.
Further does the patch ensure freeing memory with gcc-7.

The testcases have been actived again for annoucing their passing.

Reffing #293

### DIFF
--- a/src/mpi/mpi_caf.c
+++ b/src/mpi/mpi_caf.c
@@ -464,12 +464,12 @@ PREFIX (finalize) (void)
 	  MPI_Win_free (mpi_token->desc);
 	  free (mpi_token->desc);
 	}
-#else
+#else // GCC_GE_7
 # ifndef CAF_MPI_LOCK_UNLOCK
       MPI_Win_unlock_all(*p);
 # endif // CAF_MPI_LOCK_UNLOCK
+#endif // GCC_GE_7
       MPI_Win_free(p);
-#endif
       free(tmp_tot);
       tmp_tot = prev;
     }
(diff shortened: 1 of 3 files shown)


rouson:
@vehre Thanks for working on this.  It appears that this pull request causes the [syncimages_status](https://github.com/sourceryinstitute/opencoarrays/blob/vehre/events-regression/src/tests/unit/sync/syncimages_status.f90) test to hang as Travis-CI reports a the bottom of the text of [Job 945.6](https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/187506483), where the aforementioned test times out after not responding for 10 minutes.  I need to review the standard in order to understand [line 14](https://github.com/sourceryinstitute/opencoarrays/blob/vehre/events-regression/src/tests/unit/sync/syncimages_status.f90#L14) in the test better.  I haven't made much use of `sync images(*...` so I'm not that familiar with its effects.

rouson:
The [GitHub History page](https://github.com/sourceryinstitute/opencoarrays/commits/vehre/events-regression/src/tests/unit/sync/syncimages_status.f90) for the sync_images test shows two contributors: @zbeekman and me with me first merging it in from afanfa-master branch.  I'm not sure I'm understanding the history, but maybe @fanfa wrote it, I merged it into the master branch, and @zbeekman made minor white-space edits. 

At any rate, I think the the test is incorrect.  Section 11.6.4, paragraph 3 of the draft Fortran 2015 standard states, "An image-set that is an asterisk specifies all images in the current team."  Given that we don't yet support teams, I believe this quote implies that `sync images(*...` is equivalent to `sync all`, which might explain why it hangs because all images other than image 1 are waiting for image 1 to execute a corresponding synchronization (presumably `sync images(*...`).  I'll rewrite the test and see if that fixes the problem.  It would also be nice to eliminate the non-standard `sleep(1)` for purposes of supporting  compilers that don't offer that extension.  The second code listing [here](http://stackoverflow.com/questions/6931846/sleep-in-fortran) provides what I assume is a standard-conforming Fortran implementation of sleep.  For now, I'll include it as an internal program in the test, but we at some point should probably collect a few standard utilities like this into a library for use throughout the test suite (another candidate would be the assert procedure used in the [async_comp_alloc](https://github.com/sourceryinstitute/opencoarrays/blob/vehre/events-regression/src/tests/unit/init_register/async_comp_alloc.f90) test.

vehre:
When I examine the changes my patch has on non gcc-7 compiled executables, I see no change at all. The only change is in line 471 of mpi_caf.c where the MPI_Win_free(*p) is moved out of the conditional. But that statement was always executed on non-gcc-7 compiled executables, because it is on the else path of the #if GCC_GE_7 preprocessor conditional.

I could not find a F2015 standard that has a section 11.6.4, so I assume you mean 8.6.4 as in the F2015 standard from https://gcc.gnu.org/wiki/GFortranStandards. At least there is the same formulation as cited. 

Nevertheless do I disagree, that the test in syncimages_status.f90 is wrong. The test is to check whether calling sync_images() on all images with at least one image stopped succeeds in reporting the stat-code for stopped images. This of course is a race condition. When image 1 is slow to exit (due to load on the machine), the test randomly will fail. An attempt to fix this is made by #299.

rouson:
@vehre Section 11.6.4 is on page 208 (which is the 228th page in the PDF document) of the August 31 draft located [here](http://j3-fortran.org/doc/meeting/211/16-007r2.pdf).  Clearly I was mistaken in my interpretation of of the behavior of `sync images` (which I've used less often than `sync all`) and `stat_stopped_image` (which I've never used).   I agree the test is correct.  At some point soon, I'll still replace the non-standard `sleep` call in the test.

I received a new CodeCoverage message a few minutes ago, which I interpret as meaning that your commit ([84b4e7](https://github.com/sourceryinstitute/opencoarrays/commit/84b4e7f6e897fa818ea2a8e044c67813af6e7410)) launched a new round of checks, but I can't see that the Travis-CI tests are running and  I don't know how to l launch them by hand.  If the failure is intermittent due to load, then hopefully they'll pass this time.  Let's see.

rouson:
@vehre I see now that your latest commit removes the sleep. Thanks.

vehre:
Thanks for the pointer to the most recent F2015 standard.

zbeekman:
Superseded by #302

PR #299, opened 2016-12-30
Fix sync image race/hang for syncimage_status

vehre:
Please see the commit message for a detailed explanation.


vehre:
Well, at first it looked like pull request #300 was causing random failure on MacOS X for the co_heat testcase, but now it shows here also.
Therefore, please DO NOT MERGE until I have fixed this reliably. Working on it.

zbeekman:
@vehre Just wondering what the status of this is. It looks like you're still working on it, no?

vehre:
Please close and delete the branch: vehre/sync_image_fix, when satisfied with #302.

PR #302, opened 2017-01-09
Fix sync images race/hang (fixes #298, supersedes #299)

vehre:
This pull request fixes the issue open in #298.

This pull request supersedes #299, which can be closed now.


zbeekman:
@vehre please confirm that I have correctly understood you:

> This pull request fixes the issue open in #298.

OK... but #298 is a pull request... so I take it this PR can supersede #298? If the commit SHAs match for the common commits, I'll just close #298 and use this PR instead...

> This pull request supersedes #299, which can be closed now.

OK.... so I'll close #299 then...

vehre:
Well, this pull request contains or rather depends on/extends the code #298. In the discussion of #298 an issue is raised, which is now fixed by this pull request. When you merge, make sure, that you merge both, the changes to pull of #298 (which is https://github.com/sourceryinstitute/opencoarrays/pull/302/commits/0696306beb8afd42a98f570aebe6b2a295445554) and this pull request.

zbeekman:
> When you merge, make sure, that you merge both, the changes to pull of #298 (which is 0696306) and this pull request.

OK, perfect. 0696306 is on both vehre/events-regression and vehre/sync_images_fix_new. Since the tests pass here, but not on #298 I'm going to close #298 in favor of this PR. In addition I'm going to squash 4164cba and 86d57af (or drop 86d57af and amend 4164cba so that the travis notifications are never disabled.

src/tests/unit/sync/syncimages_ring.f90
```diff
@@ -0,0 +1,25 @@
+! SYNC IMAGES([this_image - 1, this_image + 1]) with the STAT=STAT_STOPPED_IMAGE
+! specifier and wrap around of image numbers.  The test is intended to check
+! that syncing in a ring with a stopped image still terminates all images.
```
zbeekman: @vehre: Great test idea! Can you please rename the test to better reflect its purpose? Right now it sounds as though it is meant to test if a ring correctly syncs. A better name might be something like `abort_sync_on_stopped_image` or something more indicative of it's purpose. You can just add one or more commits to vehre/sync_images_fix_new and push them, it will update this PR.

Also can you please add an extra `!` and move these comments below `program ...` that way the documentation will get picked up by [FORD](https://github.com/cmacmackin/ford).

src/tests/unit/sync/syncimages_ring.f90
```diff
@@ -0,0 +1,25 @@
...
+  if (num_images() .lt. 2) error stop "Need at least two images to test."
+
+  associate (me => this_image())
+    if (me /= 1) then
+	  associate (lhs => merge(me - 1, num_images(), me /= 1), &
+				 rhs => merge(me + 1, 1, me /= num_images()))
+		sync images([lhs, rhs], STAT=stat_var)
+		! Only on image 2 and num_images() testing whether a stopped image is
+		! present can be done reliably. All other images could be up ahead.
+		if (stat_var /= STAT_STOPPED_IMAGE .and. me == 2) error stop "Error: stat_var /= STAT_STOPPED_IMAGE: "
```
zbeekman: Can we please also test `num_images()` in addition to `2`? That way we will test the LHS and RHS of the stopped image, image 1. Also if you add an extra `!` to these comments then FORD will pick them up.

src/tests/unit/sync/syncimages_ring.f90
```diff
@@ -0,0 +1,25 @@
...
+program sync_images_ring
+  use, intrinsic:: iso_fortran_env
+  implicit none
+
+  integer :: stat_var = 0
+
+  if (num_images() .lt. 2) error stop "Need at least two images to test."
+
+  associate (me => this_image())
+    if (me /= 1) then
```
zbeekman: I'm not 100% confident in the semantics of calling `stop` (**NOT** `error stop`) from a coarray program, but if you can call `stop 0` from image `me == 1` explicitly without causing all the other images to abort, that would clarify the test.

src/tests/unit/sync/syncimages_status.f90
```diff
@@ -10,7 +10,6 @@ program sync_images_stat
   me = this_image()
 
   if (me /= 1 ) then
-     call sleep(1)
```
zbeekman: So if I'm understanding correctly, all the other images will wait for a response from image 1, but image 1 isn't included in the call to `sync images()` they would wait forever. However, image 1 will reach program termination, and then the other images waiting via `sync images()` will receive a STAT_STOPPED_IMAGE as the response. This is why there is no longer a race condition and/or hang in this test: all the images will be waiting for each other except image 1 which will continue until it terminates, at which point the other images will get the STAT_STOPPED_IMAGE notification. Is my understanding correct?

CMakeLists.txt
```diff
@@ -391,6 +391,9 @@ if(opencoarrays_aware_compiler)
   add_mpi_test(duplicate_syncimages 8 ${tests_root}/unit/sync/duplicate_syncimages)
   add_mpi_test(co_reduce 4 ${tests_root}/unit/collectives/co_reduce_test)
   add_mpi_test(syncimages_status 32 ${tests_root}/unit/sync/syncimages_status)
+  add_mpi_test(syncimages_ring_np3 3 ${tests_root}/unit/sync/syncimages_ring)
+  add_mpi_test(syncimages_ring_np13 13 ${tests_root}/unit/sync/syncimages_ring)
+  add_mpi_test(syncimages_ring_np23 23 ${tests_root}/unit/sync/syncimages_ring)
```
zbeekman: How did you arrive at 13 and 23?

src/tests/unit/sync/syncimages_ring.f90
zbeekman: I can do this.

src/tests/unit/sync/syncimages_ring.f90
zbeekman: I can test and implement this

src/tests/unit/sync/syncimages_ring.f90
zbeekman: I can rename it and adjust the comments

zbeekman:
Pending confirmation from @vehre that my understanding is correct, and some small tweaks, this

looks good to me

src/tests/unit/sync/syncimages_ring.f90
vehre: The motivation of the name is simple: It is testing the intrinsic statement `sync image` which should be modelled in the test's name. It is modelling a ring. That there is an abort to be expected can be reflected in the name: sync_image_ring_abort_on_stopped_image

Idea: A test sync_image_ring should be modelled @rouson where the abort is not expected and it is just tested whether a ring is syncing.

The code is just a copy from syncimage_status and I took over that style. If you want to polish the comments go ahead. Then I concentrate on fixing issues and adding functionality. It is always a good idea when the tests are not written by the dev who implemented a feature, because those tests tend to comply, because the dev knows what to do.

src/tests/unit/sync/syncimages_ring.f90
vehre: What? There is `error stop "Need ..."` called. What do I not get here?

src/tests/unit/sync/syncimages_status.f90
vehre: Nearly correct. Only the reason for the race is not absolutely correct, but a little bit more difficult to explain. In the former implementation the status of the image was read using one-sided communication. When that read was before the aborting image set its abort-status, then the program would hang. When the set of the abort-status was interleaved with some images having read the status already and some not, then the race begun, that the images that had the abort received set their own abort-status while the other images where reading that status. So timing was everything here.

Now the abort-status is send using two-sided communication, which also means, that the sync is done at this stage. That way no race is possible anymore, because an aborted-image would publish its state and all other receive it a defined location.

CMakeLists.txt
vehre: Arbitrary, in fact, just by adding a 1 and a 2 before the 3 of the initial test. I did not want to have a power of two here.

src/tests/unit/sync/syncimages_ring.f90
zbeekman: Just to make it more clear that image 1 is expected to reach termination before the other images since it is not part of the `sync images(*,...)` call. It might be a dumb/infeasable idea, I'll test it and see what happens. Otherwise I'll just add some additional comments to clarify.

src/tests/unit/sync/syncimages_ring.f90
zbeekman: > That there is an abort to be expected can be reflected in the name: sync_image_ring_abort_on_stopped_image

Yes, I meant rename the test source file.

>Idea: A test sync_image_ring should be modelled @rouson where the abort is not expected and it is just tested whether a ring is syncing.

This is a good idea. I'm trying to think of a good way to test this. My current thinking is that we could space images equally along a domain, say [0, 2pi), have each compute a periodic function (e.g. sine) at their location, and also perform a put to the left and right of what their neighbors value should be. (puts are asynchronous so the ring sync will actually be tested...) Then we do the ring sync and verify that the value put from the lhs neighbor and rhs neighbor match the local value (to within some suitable epsilon).

> It is always a good idea when the tests are not written by the dev who implemented a feature, because those tests tend to comply, because the dev knows what to do.

This is very true. Also tools like quick check that can randomly generate tests to check the properties of the API rather than hard-coded values, which is a similar idea.

CMakeLists.txt
zbeekman: OK, primes are good, but 23 may start to get a bit slow on some systems. I may tweak these values to be smaller primes.

zbeekman:
@vehre are you done with `vehre/sync_images_fix_new`? Can I force push and re-write history on this branch to clean up this PR, or will that cause problems for you?

src/tests/unit/sync/syncimages_ring.f90
vehre: Oh, now I get you.

No, that does not work. stop_numeric() and stop_string() btw is not implemented correctly in caf_mpi. Both expect caf_finalize() to exit(0), which this function *must not* do! So until that is fixed, STOP 0 will not yield the result you expect.

zbeekman: > stop_numeric() and stop_string() btw is not implemented correctly in caf_mpi. Both expect caf_finalize() to exit(0), which this function must not do! So until that is fixed, STOP 0 will not yield the result you expect.

I think we should open an issue for this, do you agree?

zbeekman:
Just to make our lives simper, I'm not going to force push even though my OCD wants to squash 86d57af with 4164cba

vehre:
I am done with the pull request. I don't expect it need more attention. Go ahead with merging, although I did not get your latest comment.

src/tests/unit/sync/syncimages_ring.f90
vehre: yes

rouson:
This should fix #293

zbeekman:
LGTM

src/tests/unit/sync/sync_image_ring_abort_on_stopped_image.f90
```diff
@@ -11,15 +13,21 @@ program sync_images_ring
...
+       associate (lhs => merge(me - 1, num_images(), me /= 1), &
+              rhs => merge(me + 1, 1, me /= num_images()))
+         sync images([lhs, rhs], STAT=stat_var)
+         !! Only images bordering image 1 (i.e., 2 and `num_images()`) can
+         !! accurately test whether a stopped image is present. All other
+         !! images could be up ahead.
+         if (stat_var /= STAT_STOPPED_IMAGE .and. me == 2) &
+              error stop "Error: stat_var /= STAT_STOPPED_IMAGE: "
+         if (stat_var /= STAT_STOPPED_IMAGE .and. me == num_images()) &
+              error stop "Error: stat_var /= STAT_STOPPED_IMAGE: "
```
zbeekman: @vehre shouldn't this be OK too? It seems that this is failing on one Travis-CI OS X test (when code coverage flags are enabled): https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/191684163#L463 Could there be an off-by-one error somewhere that I introduced, or in another part of the library/test?

vehre:
Mmh, at first glance this looks right. I don't have access to a PC currently. So you have to wait until Monday when I return home before I can analyze this in depth.

zbeekman:
> Mmh, at first glance this looks right. I don't have access to a PC currently. So you have to wait until Monday when I return home before I can analyze this in depth.

OK, thanks for taking a look. My guess is that if we re-run it, then it will go away due to a race condition somewhere... because it's not regularly showing up. At any rate, this is a fairly low priority bug since:

 1. It often works correctly
 2. If you already have stopped images with a sync images then this is likely an exception or logic error to begin with, and shouldn't have a big impact on the correct execution of most programs

zbeekman:
Also it happens on Mac OS X even when the coverage flags aren't enabled: https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/191687068

vehre:
Jikes, that was a bug. I read the sync result using the index into the array of receive requests, where it would have been better to use the MPI_SOURCE of the status parameter. This is fixed now by pull request #313.

PR #304, opened 2017-01-10
Include CheckFortranSourceCompiles where it is needed.

vehre:
The former version was including CheckCSourceCompiles instead of the Fortran
version.  This in principle does not matter, because CheckFortranSourceCompile
is included already.  For clarity nevertheless this should be made consistent.


zbeekman:
LGTM

@vehre thanks so much for catching this copy pasta error! :-)

PR #307, opened 2017-01-13
fix gfortran compiler detection (issue 305)

LaHaine:
There was a typo in the check for the compiled fortran binary that made the check fail every time.

### DIFF
--- a/prerequisites/install-functions/find_or_install.sh
+++ b/prerequisites/install-functions/find_or_install.sh
@@ -266,7 +266,7 @@ find_or_install()
       info "$this_script: Checking whether $executable in PATH is version $(./build.sh -V gcc) or later..."
       $executable -o acceptable_compiler acceptable_compiler.f90 || true;
       $executable -o print_true print_true.f90 || true;
-      if [[ -f ./accepatable_compiler && -f ./print_true ]]; then
+      if [[ -f ./acceptable_compiler && -f ./print_true ]]; then
         is_true=$(./print_true)
         acceptable=$(./acceptable_compiler)
         rm acceptable_compiler print_true


zbeekman:
@LaHaine Good catch! Thanks so much! 💯 I'm surprised this hasn't surfaced yet... I'll have to dig around a bit to better understand why.

Fixes bug introduced in 6cea2404a442e13e0aa40439ab8d05e5a16ea725

LGTM

zbeekman:
LGTM

zbeekman:
Thank you for your contribution to OpenCoarrays, @LaHaine! :bow: :nada: 🎆

prerequisites/install-functions/build_opencoarrays.sh
```diff
@@ -13,15 +13,15 @@ build_opencoarrays()
...
     emergency "build_opencoarrays.sh: empty \${MPIFC}=${MPIFC:-} or \${MPICC}=${MPICC:-}"
   fi
-  MPIFC_show=`$MPIFC -show`
-  MPICC_show=`$MPICC -show`
-  if [[ ${MPIFC_show} != *gfortran* || ${MPICC_show} != *gcc* ]]; then
-    emergency "build_opencoarrays.sh: MPI doesn't wrap gfortran/gcc: \${MPIFC_show}=${MPIFC_show}, \${MPICC_show}=${MPICC_show}"
-  fi 
+  MPIFC_show=($($MPIFC -show))
+  MPICC_show=($($MPICC -show))
+  if [[ "${MPIFC_show[0]}" != *gfortran* || "${MPICC_show[0]}" != *gcc* ]]; then
```
zbeekman: We may want to add some more robust handling in the event that the installation location has spaces in the path: I'm not sure if the MPI wrappers will properly quote and/or escape these cases... It's tricky too, because it's possible for the compiler in the wrapper to just pull from the path (e.g., `gcc`) or be a full or relative path (e.g., `/usr/local/bin/gcc`)... One idea is to use `which`, `hash`, or `type` to determine if array element 0 is a valid executable, and if not, try to add the 1st element, test again, add 2nd etc. to handle spaces... 

Another thought: parse up until the first occurrence of `[:blank:]-` which will indicate a flag... this would be fine except, perhaps for cases where the compiler has MPI built in, like Cray... but I think there is some exceptional logic to handle this anyway.

Issue #306, opened 2017-01-13
install.sh calls gcc with the wrong path


LaHaine:
I've tested install.sh again and let it install all prequisites. In the end, it tries to compile opencoarrays, but it is using the wrong path for the just compiled gcc, /batch/test/opencoarrays/prerequisites/installations/gcc instead of /batch/test/opencoarrays/prerequisites/installations/gcc/6.1.0/bin/gcc .

LaHaine:
[install-opencoarrays.log.txt](https://github.com/sourceryinstitute/opencoarrays/files/703820/install-opencoarrays.log.txt)

zbeekman:
Ah, I see.... this is because of the greedy `%%` parameter expansion and the fact that gcc appears twice once in the path and once as the name of the binary. We need to handle/parse the output of `mpicc -show` much more carefully. I also don't think we can cavalierly just discard the binary name either since some systems are gcc-6 etc.

PR #308, opened 2017-01-13
Fix setting of C compiler var (issue 306)

LaHaine:
The wild card in the build script would eat the path component
of the installed gcc 6.1.0 build and set the var to a directory
instead of .../bin/gcc

### DIFF
--- a/prerequisites/install-functions/build_opencoarrays.sh
+++ b/prerequisites/install-functions/build_opencoarrays.sh
@@ -21,7 +21,7 @@ build_opencoarrays()
   # Set FC to the MPI implementation's gfortran command with any preceding path but without any subsequent arguments:
   FC="${MPIFC_show%%gfortran*}"gfortran
   # Set CC to the MPI implementation's gcc command...
-  CC="${MPICC_show%%gcc*}"gcc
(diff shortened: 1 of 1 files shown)


prerequisites/install-functions/build_opencoarrays.sh
```diff
@@ -21,7 +21,7 @@ build_opencoarrays()
   # Set FC to the MPI implementation's gfortran command with any preceding path but without any subsequent arguments:
   FC="${MPIFC_show%%gfortran*}"gfortran
   # Set CC to the MPI implementation's gcc command...
-  CC="${MPICC_show%%gcc*}"gcc
+  CC="${MPICC_show%%gcc *}"gcc
```
zbeekman: > The wild card in the build script would eat the path component
of the installed gcc 6.1.0 build and set the var to a directory
instead of .../bin/gcc

But after eating everything, it adds back gcc.... on my system if you add the space then CC just gets set to the entire `mpicc -show` so I don't think this is actually a viable fix... I'll have to look at #306 and dig around a bit to understand why this is causing issues for you.

Issue #243, opened 2016-11-07
Allocate not correctly synchronizing on oversubscribed cores


rouson:
EDIT by @zbeekman: GFortran [PR 78505](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78505)

The transcript below demonstrates that OpenCoarrays is not performing the synchronization required during the allocation of a coarray when the number of images exceeds the number of cores (8 images on 4 cores in the case below).  I can't think of a way to test the `allocate` without `source =` because
 
1. The coarray cannot be defined before the `allocate` and
2. A definition of the coarray immediately after the `allocate` would happen asynchronously.

The workaround is to add some form of synchronization (e.g., `sync all`) after the `allocate`, but requiring that does not conform with the standard.

```
$ cat source-allocation-not-syncing.f90 
 implicit none
 integer, allocatable :: f(:)[:]
 integer, parameter :: num_points=32
 integer :: me,ni,my_num_points,neighbor_last_element
 me = this_image()
 if (mod(num_points,num_images())/=0) error stop "num_points not evenly divisible by num_images()"
 my_num_points = num_points/num_images()
 allocate( f(my_num_points)[*], source = 1 )
 if (me>1) then 
   neighbor_last_element = f(my_num_points)[me-1]
   if (neighbor_last_element /=1) print *,"Image ",me," gets ",neighbor_last_element
 end if
end 

$ caf source-allocation-not-syncing.f90 
$ cafrun -np 8 ./a.out
 Image            5  gets            0
$ cafrun -np 8 ./a.out
 Image            3  gets            0
 Image            5  gets            0

$ caf --wraps
caf wraps CAFC=/opt/mpich/3.1.4/gnu/7.0.0/bin/mpif90

$ /opt/mpich/3.1.4/gnu/7.0.0/bin/mpif90 --version
GNU Fortran (GCC) 7.0.0 20160904 (experimental)
...
```

rouson:
FYI, I believe the gfortran above gfortran date (20160904) precedes the first commit in support of allocatable components and I observe the same behavior with gfortran 6.2.0 so I suspect this behavior is not due to any recent commits.

afanfa:
If you remove the source=1 as argument you should see the correct behavior. Having the source=1 in the allocation statement gives you the illusion that the assignment happens during the allocation but actually it happens right after. That's what causes the bug.
There is nothing I can do in the library for this specific case because it is managed by the compiler.

[...]

Issue #172, opened 2016-04-16
Unexpected behavior with co_reduce


floquet:
The **co_reduce** collective appears to be posting random, large number to `result_image` using the reference example in the [gfortran documentation](https://gcc.gnu.org/onlinedocs/gfortran/CO_005fREDUCE.html). A diagnostic version of the code and results are included in Stack Overflow: [Fortran coarray anomaly with co_reduce](http://stackoverflow.com/questions/36658286/fortran-coarray-anomaly-with-co-reduce).

Sample code showing compilation and execution:
[co_reduce.tar.gz](https://github.com/sourceryinstitute/opencoarrays/files/222403/co_reduce.tar.gz)

afanfa:
Thanks for reporting this. I've found the reason for this problem and I'm producing a patch (for OpenCoarrays). What I don't completely understand is why this problem hasn't come out earlier.

rouson:
Presumably it hasn't appeared sooner because our current co_reduce unit test doesn't use the optional result_image argument.  I'll add tests that include the result_image argument.

afanfa:
Unfortunately result_image is not the problem. The patch that solves this bug breaks the co_reduce test already part of the OpenCoarrays test suite.

rouson:
Wow.  Thanks for catching this.  So if the patch breaks the current co_reduce test, does that mean the current co_reduce test is incorrect.

afanfa:
The compiler does something strange on the co_reduce test included in the test suite. I'm trying to understand why...

afanfa:
I created a new branch called bugfix-coreduce that includes the patch. I think this is the right way to do it but I don't have enough time to investigate further. The problem with the co_reduce test case is probably due to a compiler bug.

floquet:
Thanks for making coarrrays a reality. Really appreciate the hard work.

floquet:
P.S. I hit the bug a about two months ago, but was convinced the problem
was my programming or my machine.

rouson:
@afanfa, any updates?

afanfa:
The problem is mainly in the compiler. OpenCoarrays won't be affected by the change

afanfa:
No updates. The issue is related to a compiler bug.

rouson:
@afanfa, in that case, I assume it's for someone else to work on under contract.  If so, we'll try to contract out the work to someone.  If I'm incorrect and it's something you have time to work on, let us know.  No pressure.

[...]

PR #310, opened 2017-01-14
Small cleanups

### LINKED ISSUE #79: Review installation logic, files and directories

### LINKED ISSUE #297: libfft_sse.a fails to link on `mpich-docker-image`
Attn: @rouson 

When trying to build the naiver-stokes integration test, the linker fails with the following output:

```
/usr/bin/gfortran   -Wl,-rpath  -Wl,/opt/mpich/3.1.4/gnu/6.2.1/lib  -Wl,--enable-new-dtags  -g CMakeFiles/coarray_navier_stokes.dir/coarray-shear_coll.F90.o  -o coarray_navier_stokes -Wl,-rpath,/opt/mpich/3.1.4/gnu/6.2.1/lib  ../../../../../../src/tests/integration/pde_solvers/navier-stokes/libfft_sse.a ../../../../mpi/libcaf_mpi.a /opt/mpich/3.1.4/gnu/6.2.1/lib/libmpi.so /opt/mpich/3.1.4/gnu/6.2.1/lib/libmpifort.so /opt/mpich/3.1.4/gnu/6.2.1/lib/libmpi.so /opt/mpich/3.1.4/gnu/6.2.1/lib/libmpifort.so
/usr/bin/ld: ../../../../../../src/tests/integration/pde_solvers/navier-stokes/libfft_sse.a(trig.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../../../../../src/tests/integration/pde_solvers/navier-stokes/libfft_sse.a(plan.o): relocation R_X86_64_32 against `.rodata.str1.4' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
```

I'm a bit perplexed as to why this happens... the linker is saying that `-fPIC` should be added, but, I thought that the FFT libraries were hand written in assembly... Perhaps this is just a "won't fix" and we should move on...

zbeekman:
This branch name has become a bit of a misnomer... I have lumped together a number of fixes and enhancements... in general this is probably a bad practice, but I vote we let it slide for now...

Add badges etc to readme, adjust regression tests links to reflect closure of issues, and add co_reduce regression test, even though it appears it was never a regression to begin with.

 - Fixes #79
 - Fixes  #297 
 - Add regression test for #243 
 - Add regression for #172 (currently set to pass when test fails)
 - Migrates to use of a `.VERSION` file to make parsing easier for scripts and allow extra comments
 - Adds auto-upload of release assets upon tagging with git, but requires that the tag is PGP signed (`git tag -s <tag> [tree-ish]`)
 - This should compute the SHA 256 checksum and create a detached signature of the cryptographic SHA 256 checksum with an encrypted GPG subway I uploaded to the repo/travis. 🔮 🎩 🐇


zbeekman:
These commits need to be rebased and cleaned up before merging... Also I set CMake/CTest to indicate pass status even when regression 172 fails... I did this so that I could examine and compare codecov.io output between the passing and failing tests in an attempt to localize the bug if it's in the library.

zbeekman:
With `value` instead of `intent(in)` here are the coverage changes: https://codecov.io/gh/sourceryinstitute/opencoarrays/compare/882c371d4d7e84364eb7adfba4b4f8d840e3f398...cb9f040bfc7543424a637a43fb447fbbe605a0f8/changes

vs

https://codecov.io/gh/sourceryinstitute/opencoarrays/compare/882c371d4d7e84364eb7adfba4b4f8d840e3f398...1a0e3b7edb6867d9e9371cbd9ed1d8f5b3dd2010/changes with `intent(in)`

zbeekman:
lgtm

zbeekman:
Need to account for older CMake where `CMAKE_Fortran_COMPILER_VERSION` is missing....

zbeekman:
LGTM

zbeekman:
@rouson: I've added a regression test for #172, but it's still failing. To cope with this, I've set it so that failing indicates success (so that users won't see test failures because it's a known issue...) I'd like to merge this PR, but I could split that test out into a separate one. What do you think? Would you be OK with merging this PR with the failing regression for #172 but have it report a "PASS" status when it fails?

zbeekman:
I'm going to go for it... we can always factor out the #172 regression test later.

vehre:
Is there no way to set a test to expect to fail?

zbeekman:
That's what it is now. But then when it's fixed it will report a failure.
It's a logical not/inverse operation.

rouson:
@zbeekman Hopefully you saw my message sent via Hangouts.  If a failing test degrades user confidence, then I believe that a false positive would degrade user confidence even more.  Moreover, I am highly unlikely to remember this exchange 6 months from now when a user reports a related problem and then I spend time trying to track down the cause, only to realize the cause has been under my nose all along, but I forgot about it because I've ben lulled into a false sense of security.   Probably it's best to set up CMake so that there are a group of tests that are cordoned off as known failures and not run unless a flag is passed to run them.  They should probably also be listed somewhere such as in our [Known Issues](https://github.com/sourceryinstitute/opencoarrays/blob/master/STATUS.md) page, which actually might be obsolete at this point because we haven't maintained it.  So probably an even better approach would be to have a Known Issues milestone under which we would the issue numbers related to failing tests and name the tests.  I'd say almost anything is better than having a test falsely report success.

zbeekman:
This is clearly the correct way to do it. The only question is: when do we turn on the known failures? During CI tests? For the users by default (with a message telling them it's a known deficiency?) I'm inclined to take the latter approach, in the interest of transparency 

As for the known issues page, we should ensure that we have an open issue for each issue and a link to the issues page.

rouson:
@zbeekman Your suggestion sounds good to me as long as the users see a message indicating that these are known failures.  Ideally the message would even include a URL to the list of known failures.  For Travis-CI, we would build with something like

cmake .. -DBUILD_KNOWN_FAILURES=TRUE

where the above flag would default to FALSE or would be explicitly set to FALSE for CI.

zbeekman:
> Your suggestion sounds good to me as long as the users see a message indicating that these are known failures.

I 100% agree... My thinking is this: In the known failures test we output clear messages with URLs pointing to the issue tracker pages saying these are known deficiencies that we are working to resolve. Failing tests are turned on by default, and if the user is installing via `install.sh` we can ensure that we `export CTEST_OUTPUT_ON_FAILURE=1` and have any calls to `ctest` with the `--output-on-failure` flag. This way they will see the warning message. We can also include a warning in the CMake configure script informing them that we have turned on tests with known failures and giving them instructions on how to turn those off.

As long as the failures are runtime failures we should always build them, since compilation in and of itself is a smoke test.

On Travis-CI, however, I don't think we should run tests known to fail so that we can continue our workflow of finding new bugs and issues, without the noise coming from bugs we already know about. Otherwise the known issues would obfuscate new ones and coding errors/typos etc.

Issue #268, opened 2016-12-04
RFC: Hard code full paths and more precisely specify linker/compiler flags & MPI flags in caf and cafrun scripts


zbeekman:
Request for comment: wrapper script re-engineering

`caf` is to `mpif90` as `cafrun` is to `mpirun`: they are wrapper scripts for compiling and launching OpenCoarrays enabled coarray fortran programs. Currently the scripts use a bit of an ad-hoc sandwich technique to get written: Some static content exists in a "head" file, some static content exists in a "foot" file and then some content is echoed during the build process to get inserted into the middle.

My proposal is:
 1. Use CMake's native `configure_file` capabilities to store a `caf.in` and a `cafrun.in` file and use CMake's `@variable@ expansion to substitute the correct values into this file.
 2. Specify explicitly the full paths to the GCC (or, perhaps in the future, a different) Fortran compiler, rather than using the MPI wrapper script.
 3. Specify explicitly the compiler and linker flags needed for MPI via the FindMPI cmake module
 4. Allow backwards compatibility fallback to the compiler wrapper scripts when the build detects the user has explicitly passed the MPI wrappers as the C and Fortran compilers.
 5. This behavior of passing `FC=mpif90` etc. will be deprecated (but temporarily, unofficially supported for the medium term) to ease package maintainers and users jobs who have become accustomed to this behavior.

Thoughts, @rouson, @jerryd, @afanfa, @vehre, and anyone else?

zbeekman:
One further concern of mine is how much effort or duplicate effort this may be on the caffeinated opencoarrays branch.

vehre:
@1: The way it should be.
@2: Think about huge computers, that have more then 5 different compilers installed (like e.g. NERSC). Will you be going to hardcode the paths for every compiler? Idea: Use a default and allow the paths to compilers, libraries, includes and so on be overidden by environment variables that are set by the module system controlled by the admins accordingly. 
@3++: When allowing the override by the module system, then the flags need to be overidable, too.

rouson:
@vehre 

Zaak's point is to capture what was used during the installation process.  That's the only thing that is expected to work.  Switching to a different compiler absolutely will not work because OpenCoarrays currently only supports one compiler: gfortran.  We definitely want to support other compilers and have taken some steps toward doing so but that requires tremendous resources that are not currently available.  In our current state, overriding anything will generally lead to failure.

[...]

PR #311, opened 2017-01-15
Add host assoc event test

rouson:
This pull request adds a test that exposes the internal compiler error described
in GCC [bug report 70696](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70696).  
Because the test currently generates a compile-time error, A CMake macro is
added to delay the building of the test until the ctest runtime.


zbeekman:
@rouson Besides the issue that @vehre and I are already aware of with the ring-abort test, it seems that CTest cannot find a script for the new test you added:

https://travis-ci.org/sourceryinstitute/opencoarrays/jobs/192138710#L483

src/tests/unit/events/host-associated-alloc-comp.f90
```diff
@@ -0,0 +1,43 @@
+! host-associated-alloc-comp.f90
+!
+! -- Demonstrate an internal compiler error (ICE) on EVENT POST of host-associated EVENT_TYPE coarray 
+!    gfortran Bug Reprot 70696 (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70696)
```
zbeekman: Small typo: Reprot --> Report

src/tests/unit/events/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+
```
zbeekman: Is this just a place holder file?

CMakeLists.txt
```diff
@@ -431,3 +431,6 @@ endif()
 
 include(cmake/AddInstallationScriptTest.cmake )
 add_installation_script_test(installation-scripts.sh src/tests/installation/)
+
+include(AddCompileTimeTest.cmake)
```
zbeekman: Can we please add new cmake modules to the top level `cmake` subdirectory, with the other ones? It just helps keep the top level project more organized and less cluttered IMO, but I'm open to arguments that contradict this point of view.

AddCompileTimeTest.cmake
```diff
@@ -0,0 +1,37 @@
...
+    COPYONLY
+  )
+  # Write a script to compile the code at test time instead of at build time so 
+  # CMake doesn't bail during the build process because of the compilation failure
+  set(bug_harness "${CMAKE_BINARY_DIR}/staging/test-${base}.sh")
+  install(
+      FILES "${bug_harness}"
+      PERMISSIONS WORLD_EXECUTE WORLD_READ WORLD_WRITE OWNER_EXECUTE OWNER_READ OWNER_WRITE GROUP_EXECUTE GROUP_READ GROUP_WRITE
+      DESTINATION ${CMAKE_BINARY_DIR}/${path}
+  )
```
zbeekman: The install command doesn't happen at test time... usually it happens after tests are run: people usually run tests to verify the build before actually installing it on their system. I think that we may need to use CMake in script mode to achieve what you're attempting to do here... I need to do some more research about this, however.

AddCompileTimeTest.cmake
```diff
@@ -0,0 +1,37 @@
...
+  # Copy the source to the binary tree 
+  configure_file(
+    ${CMAKE_SOURCE_DIR}/${path}/${base}.${ext} 
+    ${CMAKE_BINARY_DIR}/${path}/${base}.${ext} 
+    COPYONLY
+  )
+  # Write a script to compile the code at test time instead of at build time so 
+  # CMake doesn't bail during the build process because of the compilation failure
+  set(bug_harness "${CMAKE_BINARY_DIR}/staging/test-${base}.sh")
+  install(
```
zbeekman: You can use the [`file` command](https://cmake.org/cmake/help/v3.7/command/file.html) with the `COPY` or `INSTALL` signature. This will happen at CMake/configure time. It also allows you to set the permissions, etc.

zbeekman:
@rouson Oh man, this is a bit of a pain to get working. First of all I think we need to get the writing, staging and installation of `caf` and `cafrun` to be a little bit more robust. Especially since you can use `DESTDIR` during `make` to override what you set in `-DCMAKE_INSTALL_PREFIX`. This work can be lumped in with #268.

The build system should:

 1. Configure `caf.in` and `cafrun.in` and deploy them to `bin_staging` at compile time, with correct executable permissions.
 2. `bin_staging` scripts should point to libraries, module files, etc. in the build tree, NOT the final install path
 3. Reconfigure `caf.in` and `cafrun.in` and deploy them to the installation directory, `<prefix>/bin` with correct permissions.
 4. These installed wrappers should point to the installed libraries, headers, and module files, preferably using a relative path, in the event that someone wants to attempt to pick up the entire directory and move it.

I'll copy this last bit into #268

zbeekman:
@rouson, we should investigate this again.... I may close this PR in the mean time, since it's quite stale. But we should come back to this, it would be nice to add some tests for known bugs/annoyances upstream to help keep us informed as to the state of the compiler.

rouson:
I agree wit closing this.  The compiler bug that inspired this has been fixed.

PR #313, opened 2017-01-16
Fix sync images to take the syncer's id from mpi_status.

vehre:
Take the source index of the image that is sending the current image data from
the MPI_status handle of Waitany instead of the index i, which only denotes the
handle in the array of receive-requests.

Fixes #302


zbeekman:
Great! 💯 

Assuming that the tests indicate that this has been resolved, it

looks good to me!

Once the tests pass, I will merge this ASAP

PR #323, opened 2017-01-31
Fixes #321 bad paths in wrapper scripts

### LINKED ISSUE #321: Incorrect setting for caf_lib_dir in Installed caf script
Using a cmake installation prefix in `/opt/...`, I found that the installed `caf` script sets `caf_lib_dir=lib64` which is incorrect.  It should have been prefixed by the value of `CMAKE_INSTALL_PREFIX`. You're getting the value from `CMAKE_INSTALL_LIBDIR`  which is set by the `GNUInstallDirs` module, and it isn't always an absolute path.  I think maybe you want `CMAKE_INSTALL_FULL_LIBDIR` here.

Same issue for `caf_mod_dir`, but that was irrelevant in my case.

I think this is likely related to some of the issues from #268.

zbeekman:
- N.B. I think things will break if the user passes `DESTDIR=<...>
   make install` This needs to be made more robust

 - CMake should write one script via `configure_file()` at build time
   into a staging directory that points to the build locations of
   everything and is made a+x

 - CMake should then use `configure_file()` again during `make
   install` so that it can properly handle cases where the user might
   set `DESTDIR=`

#268 should attempt to address this.


zbeekman:
LGTM

at least once Travis-CI for OS X catches up...

PR #326, opened 2017-02-04
Add Sourcery Institute logo to README

rouson:


zbeekman:
@rouson FYI I set up PullApprove so that changes to `.md` files don't need explicit review... So, here is a gratuitous

LGTM

PR #328, opened 2017-02-07
Edit INSTALL.md to match current options

rouson:


INSTALL.md
```diff
@@ -7,116 +7,135 @@
...
+Package management options for macOS (formerly known as OS X), Windows, and Linux are described first 
+below. Also described below are options for installing via the Sourcery Institute virtual machine or
+the OpenCoarrays installation script.
+
+### macOS ###
 
-As of release 1.2.0, users might consider installing by downloading and uncompressing
-a file from our [Releases] page and running the installation script in the top-level
-source directory:
+OS X users may use the [Homebrew] or [MacPorts] package management systems to install OpenCoarrays.
```
zbeekman: Can you please add a note that Homebrew is the preferred method? I should be able to get it to the point where our CI/release scripts automatically submit a PR with the updated Homebrew formula to homebrew-core. Then we won't have to continue maintaining MacPorts, unless you really want to...

INSTALL.md
```diff
@@ -7,116 +7,135 @@
 Installing OpenCoarrays
 =======================
 
-[![Download as PDF][pdf img]](http://md2pdf.herokuapp.com/sourceryinstitute/opencoarrays/blob/master/INSTALL.pdf)
+[The following link is commented until it can be updated. To uncomment delete the backslashes, the ourtermost square brackets, and the trailing colon and hash mark]:#
+[\[\!\[Download as PDF]\][pdf img]](http://md2pdf.herokuapp.com/sourceryinstitute/opencoarrays/blob/master/INSTALL.pdf)]:#
```
zbeekman: @rouson Please don't comment this out: This is a service that I setup that automatically generates the PDFs from the markdown pages when they are posted to GitHub, so it is never out of date! The best thing to do, assuming that 51a047f53110b0c28f944ec0b2c917c283b9b84c is still the HEAD commit on `edit-install-doc` would be the following:

```bash
git checkout edit-install-doc
git reset --hard ^51a047f53110b0c28f944ec0b2c917c283b9b84c # drop this commit by resetting to its parent (thats what the ^ carat indicates)
git push --force origin edit-install-doc
```

This will delete all commits after 51a047f53110b0c28f944ec0b2c917c283b9b84c's parent commit (so don't do this if you have committed more local changes) and reset HEAD to be 51a047f53110b0c28f944ec0b2c917c283b9b84c's parent commit, i.e., 28054a6c7e75e5de756a9f2f107853792d764883

zbeekman:
Also, in the future if you put `[ci skip]` in your commit comments when you're only editing markdown files then we won't have to run all these commits through Travis-CI

INSTALL.md
zbeekman: Also, this fails to correctly comment out the link: https://github.com/sourceryinstitute/opencoarrays/blob/51a047f53110b0c28f944ec0b2c917c283b9b84c/INSTALL.md

rouson: I reverted the commit that attempted to comment the PDF link.  When I tested it, I forgot that it points to the master branch and therefore would not reflect my changes on a separate branch so I mistakenly thought something had to be done to update the PDF.  Thanks for clarifying.

INSTALL.md
rouson: I just added the following "We recommend [Homebrew]."  As long as afanfa is willing to continue maintaining the MacPorts port, I'd like to keep it as an option.  Even if Homebrew is better, switching is non-trivial and therefore presents a barrier to existing MacPorts users.

rouson:
@zbeekman I think I addressed each of your concerns except that I forgot to include the [ci skip] text in my post recent commit.  I see links labeled "See review" and "Dismiss review", but I don't see any way to indicate that I've made the changes requested in the review so I hope it's ok if I go ahead and merge this branch into master and delete the branch.

zbeekman:
> @zbeekman I think I addressed each of your concerns except that I forgot to include the [ci skip] text in my post recent commit. I see links labeled "See review" and "Dismiss review", but I don't see any way to indicate that I've made the changes requested in the review so I hope it's ok if I go ahead and merge this branch into master and delete the branch.

Good enough for government work! The usual protocol is for the person who did the review to sign off on it, but these are small documentation changes...

LGTM!

PR #392, opened 2017-05-25
Adding developer-scripts/hpclinux-install.sh

rouson:
This script has not been tested in its fully-integrated form, but each step was tested manually on HPCLinux.  Something similar should work on most Fedora-based distros, but it's probably better to instead diagnose the problems with the likely cause (stack.sh) on a newer distribution downstream of Fedora. A similar failure was observed on CentOS 6.  I'll investigate this in a virtual machine at some point and create a related issue and fix when I have a better handle on what the specific failure mode is.


zbeekman:
:shipit:

PR #510, opened 2018-02-23
Support F2018 QUIET specifier for STOP and ERROR STOP statements

jabl:
Fortran 2018 adds a QUIET specifier to the STOP and ERROR STOP
statements, allowing the programmer to suppress the output of
signaling FP exceptions and the stop code.

This patch adds the necessary changes to OpenCoarrays due to the
GFortran changes to handle this specifier. See GFortran PR 84519. The GCC changes are in r257928.

This pull request (PR) is a:

- [ ] Bug fix
- [ ] Feature addition
- [x] Other, Please describe:

### I certify that ###

- [x] I reviewed and followed the [contributing guidelines], including
      - Increasing test coverage for all feature-addition PRs
      - Increasing test coverage for all bug-fix PRs for which there
        does not already exist a related test that failed before the PR
      - At least maintaining test coverage for all other PRs
      - Ensuring that all tests pass when run locally 
      - Naming PR  to indicate work in progress (WIP) and to attach the PR
        to the appropriate bug report or feature request [issue]
      - White space (no trailing white space or white space errors may
        be introduced)
      - Commenting code where it is non-obvious and non-trivial
      - Logically atomic, self consistent and coherent commits
      - Commit message content
      - Waiting 24 hours before self-approving the PR to give another


src/mpi/mpi_caf.c
```diff
@@ -2699,7 +2700,7 @@ PREFIX (sendget) (caf_token_t token_s, size_t offset_s, int image_index_s,
           char error_str[error_len];
           strcpy (error_str, "MPI-error: ");
           MPI_Error_string (mpi_error, &error_str[11], &error_len);
-          PREFIX (error_stop_str) (error_str, error_len + 11);
+          error_stop_str (error_str, error_len + 11, false);
```
zbeekman: @jabl Hi Janne, don't we need to keep the prefix macro here? Did something change on the GFortran side that makes this no longer the case?

jabl: I made a new static function that the PREFIX versions wrap. Since it's used internally as well I thought that would be the best way to handle the API difference due to the new quiet argument.

zbeekman: Ah, I see now. So the PREFIX version calls the internal version. Thanks for clarifying.

rouson:
Thanks for this contribution, @jabl !

Issue #589, opened 2018-09-25
File pushed by mistake


afanfa:
[issue stat img]: https://img.shields.io/issuestats/i/github/sourceryinstitute/OpenCoarrays.svg?style=flat-square

| Avg response time              |
|:------------------------------:|
| ![Issue Stats][issue stat img] |

Please make sure to put any logs, terminal output, or code in
[fenced code blocks]. Please also read the [contributing guidelines]
before submitting a new issue.

# Defect/Bug Report #

Defect: File pushed by mistake

The file called AITuning_OpenCoarrays.cpp has been pushed on the repository by mistake in the directory src/mpi. This cause the makefile to fail.
That file should be removed from the repository.

## Steps to Reproduce ##

# Question 

# Request for Enhancement (RFE) #]

[links]:#
[Google]: https://www.google.com
[Stack Overflow]: https://stackoverflow.com
[fenced codeblocks]: https://help.github.com/articles/creating-and-highlighting-code-blocks/
[GFortran bug tracker]: https://gcc.gnu.org/bugzilla/
[Intel MPI]: https://software.intel.com/en-us/intel-mpi-library
[MPT]: http://www.sgi.com/products/software/sps.html
[MVAPICH]: http://mvapich.cse.ohio-state.edu
[MPICH]: https://www.mpich.org
[Open-MPI]: https://www.open-mpi.org
[CONTRIBUTING.md]: https://github.com/sourceryinstitute/OpenCoarrays/blob/master/CONTRIBUTING.md
[Helping Out]: https://github.com/sourceryinstitute/OpenCoarrays/blob/master/CONTRIBUTING.md#helping-out
[official release]: https://github.com/sourceryinstitute/OpenCoarrays/releases
[CMake install]: https://github.com/sourceryinstitute/OpenCoarrays/blob/master/INSTALL.md#cmake-scripts
[homebrew]: https://brew.sh
[issues page]: https://github.com/sourceryinstitute/OpenCoarrays/issues
[mailing list]: https://groups.google.com/forum/#!forum/opencoarrays
[install.sh]: https://github.com/sourceryinstitute/OpenCoarrays/blob/master/install.sh

zbeekman:
Fixed in PR #590

PR #590, opened 2018-09-25
Bugfix remove aituning wrapper

afanfa:
This solves #589 . @zbeekman any idea about why this fails to pass the first TravisCI test?


zbeekman:
LGTM. macOS build failing due to package updates. I am working on a partial fix but I'm going to merge and close this in the meantime.

Issue #630, opened 2019-01-18
RFE: optimize FORM TEAM by removing extra barrier


nathanweeks:
Execution of a FORM TEAM statement results in the following calls to MPI routines in mpi_caf.c:
```
void PREFIX(form_team) (...)
{
...
  MPI_Comm *current_comm = &CAF_COMM_WORLD;
...
  ierr = MPI_Barrier(CAF_COMM_WORLD); chk_err(ierr);
...
  ierr = MPI_Comm_split(*current_comm, team_id, caf_this_image, newcomm);
...
}
```
MPI_Comm_split() is effectively a barrier: no MPI processes in CAF_COMM_WORLD can return from MPI_Comm_split() until all MPI processes in CAF_COMM_WORLD have called it. It seems that the aforementioned MPI_Barrier could be removed while still maintaining the FORM TEAM synchronization semantics as specified in the Fortran 2018 (N2146 draft):

> **11.6.9 FORM TEAM statement**
> ...
> 5 ...When a FORM TEAM statement is executed, there is an implicit synchronization of all active images in the current team. On those images, execution of the segment following the statement is delayed until all other active images in the current team have executed the same statement the same number of times in this team. The segments that executed before the FORM TEAM statement on an active image of this team precede the segments that execute after the FORM TEAM statement on another active image of this team...

zbeekman:
resolved in PR #660. Thanks so much @nathanweeks!!! You rock!

PR #660, opened 2019-04-28
Implement RFE #630

nathanweeks:
Since #630 is about to be closed due to staleness, I went ahead and implemented the minor change suggested in the RFE.

## Summary of changes ##

Remove the superfluous MPI_Barrier() that is called during the execution of a FORM TEAM statement.

## Rationale for changes ##

MPI_Comm_split() alone is sufficient to provide image ordering semantics required by an image control statement like FORM TEAM.

This pull request (PR) is a:

- [ ] Bug fix
- [ ] Feature addition
- [X] Other, Please describe: optimization

### I certify that ###

- [X] I certify that:
    be introduced)

### DIFF
--- a/src/mpi/mpi_caf.c
+++ b/src/mpi/mpi_caf.c
@@ -8213,7 +8213,6 @@ void PREFIX(form_team) (int team_id, caf_team_t *team,
   MPI_Comm *current_comm = &CAF_COMM_WORLD;
   int ierr;
 
-  ierr = MPI_Barrier(CAF_COMM_WORLD); chk_err(ierr);
   newcomm = (MPI_Comm *)calloc(1,sizeof(MPI_Comm));
   ierr = MPI_Comm_split(*current_comm, team_id, caf_this_image, newcomm);
   chk_err(ierr);


src/mpi/mpi_caf.c
```diff
@@ -8213,7 +8213,6 @@ void PREFIX(form_team) (int team_id, caf_team_t *team,
   MPI_Comm *current_comm = &CAF_COMM_WORLD;
```
zbeekman: This is a bug, isn't it? Shouldn't `form team` create the new team set from the old team's communicator.

nathanweeks: It looks OK. MPI 3.1 lists the following arguments for MPI_Comm_split:

```
MPI_COMM_SPLIT(comm, color, key, newcomm)
 IN     comm                        communicator (handle)
 IN     color                       control of subset assignment (integer)
 IN     key                         control of rank assigment (integer)
 OUT    newcomm                     new communicator (handle)
```

For the "comm" argument, `*current_comm` (aka `CAF_COMM_WORLD`), which represents the current team, is used. For the last argument ("newcomm"), which stores the new communicator (team) that is stored in the FORM TEAM *team-variable* argument, `newcomm` is used. [and yes, copy & paste & resulting browser spell check revealed a typo in the MPI 3.1 standard: "assigment" -> "assignment". I'll notify a colleague who is on the MPI standards committee...]

zbeekman: Ah, I guess I find the name `CAF_COMM_WORLD` to be misleading in this context. I assumed `CAF_COMM_WORLD` would always be the communicator copied from `MPI_COMM_WORLD`, as opposed to the communicator for the present team...

zbeekman: To further elaborate: I'm concerned that a second `FORM TEAM` would partition the entire set of images, rather than the current team's set of images. But I can hunt through the source in a second to verify that `CAF_COMM_WORLD` is getting set to the communicator for the current team. (Although, if this is the case I wonder if it is a good design decision since we should still probably keep track of our copy of `MPI_COMM_WORLD` on each image.)

zbeekman:
I'm still investigating the `CAF_COMM_WORLD` stuff, but I'm going to merge this since it's orthogonal to my concerns.

src/mpi/mpi_caf.c
nathanweeks: I think the variable name CAF_COMM_WORLD predates teams support; it would be better to have it named something like CAF_COMM_CURRENT_TEAM.

The original duplicate communicator of MPI_COMM_WORLD (i.e., the 'original' CAF_COMM_WORLD) can be accessed, e.g., by searching backwards through the used_teams linked list.

PR #752, opened 2022-03-22
Use dynamic symbol resolution for libmpi_caf on macOS

zbeekman:
I'd like to get the CI system back up to parity with the old Travis-CI system. If I have time I'll add this to this PR, otherwise I'll change this from a draft to a real PR and just commit what's here.

Unfortunately mpi_caf.c now calls into libgfortran for the random seed
functionality.
It would be nice to avoid this if possible, but it's stateful so I
doubt that it would be possible.
Any code that links to the library should be Fortran code and/or pull in libgfortran.

[links]:#
[contributing guidelines]: https://github.com/sourceryinstitute/OpenCoarrays/blob/master/CONTRIBUTING.md
[issue]: https://github.com/sourceryinstitute/OpenCoarrays/issues
[coverage]: https://img.shields.io/codecov/c/github/sourceryinstitute/OpenCoarrays/master.svg?style=flat-square

|  coverage on master         |
|:---------------------------:|
| ![Codecov branch][coverage] |

## Summary of changes ##

On macOS link the caf_mpi shared dylib so that symbols from libgfortran that are unresolved when building the library can be dynamically resolved at runtime (or when the executable is linked, perhaps?).

## Rationale for changes ##

Recent changes broke OpenCoarrays on macOS. These changes were made to handle RNG seeding behavior and support make calls directly into libgfortran. Since this is stateful I doubt there is any way to avoid this, but if there is we should consider it. By default, macOS doesn't like unresolved symbols when creating shared/dynamic .dylib libraries, but you can tell the macOS linker to ignore undefined symbols and resolve them dynamically (at runtime or link time of the executable I presume). On linux the linker doesn't seem to care about creating shared libraries with undefined symbols which is why this issue didn't affect linux.

## Additional info and certifications ##

This pull request (PR) is a:

- [X] Bug fix
- [ ] Feature addition
- [ ] Other, Please describe:

### I certify that ###

- [X] I certify that:
    be introduced)

## Code coverage data

coverage on master

[links]: #
[best practices]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[commit messages]: https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message
[Pull Requests]: https://github.com/sourceryinstitue/OpenCoarrays/pulls

### DIFF
--- a/src/mpi/CMakeLists.txt
+++ b/src/mpi/CMakeLists.txt
@@ -80,6 +80,14 @@ set_target_properties ( caf_mpi
 #  VERSION ${PROJECT_VERSION}
 )
 
+# Unfortunately caf_mpi.c calls a function directly from libgfortran and this can be resolved later
+if(APPLE)
+  set_target_properties ( caf_mpi
+    PROPERTIES
+    LINK_FLAGS "-undefined dynamic_lookup"
+  )
+endif()
+
 # Create a symlink in the include dir
 if(UNIX)
   add_custom_command(TARGET caf_mpi


vehre (approved):
LGTM

I don't know what else could be done, besides replicating the code from the gfortran library and thus creating a maintenance nightmare. So when this works, go for it!

zbeekman:
> I don't know what else could be done, besides replicating the code from the gfortran library and thus creating a maintenance nightmare. So when this works, go for it!

Yes, I completely agree. Because it's stateful I don't think there's any other option.

It seems like there are a few small functions copied over from the single image implementation in libgfortran--I wonder if we should remove those, declare `extern` interfaces and pick them up from libgfortran so we're not duplicating and maintaining that code... Maybe a discussion for another time...

